PR #54, opened 2021-07-09
Restructure files and refactor imports

chocolatkey:
Files and folders have been restructured into a more standard golang layout, and the `models` package has been renamed to `pub`


mickael-menu (approved):
👌

PR #55, opened 2021-07-14
More restructuring of the server, adding internal packages

chocolatkey:
This "completes" the initial reorganization of the project.


cmd/main.go
```diff
@@ -1,426 +1,46 @@
...
+	"github.com/spf13/viper"
 )
 
-type currentBook struct {
-	filename    string
-	publication pub.Publication
-	timestamp   time.Time
-	bleveIndex  bleve.Index
-	indexed     bool
+func init() {
```
mickael-menu, line 15: We will probably add other CLI tools later on in this repo. So in the spirit of the monorepo, I suggest we add a new package `cmd/server/` or `cmd/pubserver` to contain all the stuff related to the "plug and play" publication server, including:

* A `README.md` dedicated to the publication server.
* The sample `configs/` folder and the `internal/config` package.
* `opds.go`, `server.go`, `structs.go`
    * Maybe some of the code inside could later on (not important now) be extracted in a public `pkg/` package to be used as building blocks for custom servers.

internal/config/config.go
```diff
@@ -0,0 +1,123 @@
...
+	StaticPath      string
+	Origins         []string
+
+	LogFile   string
+	LogFormat string
+	LogLevel  string
+}
+
+// NewConfig creates a Config instance
+func NewConfig() *Config {
```
mickael-menu, line 31: For "bag of data" structs without much behavior, prefer using value instead of pointer semantics. This is a bit safer by preventing an instance from being modified by an object that doesn't "own" it.

```suggestion
func NewConfig() Config {
```

server.go
```diff
@@ -0,0 +1,354 @@
...
+	"github.com/opds-community/libopds2-go/opds2"
+	"github.com/readium/r2-streamer-go/pkg/decoder/lcp"
+	"github.com/readium/r2-streamer-go/pkg/fetcher"
+	"github.com/readium/r2-streamer-go/pkg/parser"
+	"github.com/readium/r2-streamer-go/pkg/pub"
+	"github.com/readium/r2-streamer-go/pkg/searcher"
+	"github.com/urfave/negroni"
+)
+
+type R2GoServer struct {
```
mickael-menu, line 26: Maybe `PublicationServer` or `PubServer` instead of `R2Go`?

```suggestion
type PublicationServer struct {
```

internal/config/config.go
```diff
@@ -0,0 +1,123 @@
...
+	return pflag.NormalizedName(name)
+}
+
+// BindFlags normalizes and parses the command line flags
+func (cnf *Config) BindFlags() {
+	cnf.addFlags(pflag.CommandLine)
+	err := viper.BindPFlags(pflag.CommandLine)
+	if err != nil {
+		m := fmt.Sprintf("Error binding flags: %v", err)
+		logrus.Panic(m)
```
mickael-menu, line 94: I like to hide third-party dependencies with far reaching use behind a local API surface. This way if we need to change to a different dependency or upgrade to a breaking version, there's only a single place impacted.

Here maybe you could add some custom `Debugf()`, `Errorf()`, etc. (whatever you need) in the `logging` package to hide the `logrus` dependency?

mickael-menu (changes requested):

internal/config/config.go
chocolatkey: Sure thing. I do have a tendency to unnecessarily use pointers in go, so feel free to correct those.

cmd/main.go
chocolatkey: @mickael-menu is the structure in the latest commit what you are imagining? I don't really like having the cmd/server full of a lot of files, having them in the main package means you can't for example use `go run cmd/server/main.go` anymore.

mickael-menu: I would keep under `cmd/server/` only the `main()` function and its direct dependencies (initialization of the dependency graph for example) and the rest can go either:
* in `cmd/server/internal` if it's only relevant to the plug and play server.
* in a public `pkg/` package if it's a useful building block that can be reused in custom server implementations.

But as long as we have a clear separation between each command and the public packages, feel free to structure as you prefer.

internal/config/config.go
chocolatkey: Added to TODO, will be done in a separate branch/portion

PR #59, opened 2021-07-27
Refresh shared models, add Fetcher and Media Type APIs

chocolatkey:


pkg/fetcher/resource.go
```diff
@@ -0,0 +1,259 @@
...
+		return "", ex
+	}
+	xml.NewDecoder().
+}*/
+
+// Errors occurring while accessing a resource.
+type ResourceException struct {
+	Code  int
+	Cause error
+}
```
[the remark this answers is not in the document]
chocolatkey: What I ended up going with to solve this issue:
```go
type ResourceErrorCode uint16

// Error codes with HTTP equivalents
const (
	CodeBadRequest                   ResourceErrorCode = http.StatusBadGateway
	CodeNotFound                     ResourceErrorCode = http.StatusNotFound
	CodeForbidden                    ResourceErrorCode = http.StatusForbidden
	CodeServiceUnavailable           ResourceErrorCode = http.StatusServiceUnavailable
	CodeInsufficientStorage          ResourceErrorCode = http.StatusInsufficientStorage
	CodeRequestedRangeNotSatisfiable ResourceErrorCode = http.StatusInsufficientStorage
	CodeGatewayTimeout               ResourceErrorCode = http.StatusInsufficientStorage
	CodeInternalServerError          ResourceErrorCode = http.StatusInsufficientStorage
)

// The rest of the codes
const (
	_ ResourceErrorCode = iota + 1000 // Starts at 1k to not conflict with HTTP-based codes
	Offline
	Cancelled
)

// Errors occurring while accessing a resource.
type ResourceError struct {
	Cause error
	Code  ResourceErrorCode
}

func (ex ResourceError) HTTPStatus() int {
	if ex.Code > 999 { // HTTP status codes can only be three digits
		return http.StatusInternalServerError
	}
	return int(ex.Code)
}
```
I think this is the best of both worlds. Only one number needs to be stored, and there's plenty of room for numbers higher than any HTTP status code.

chocolatkey:
@mickael-menu since last time:
- the archive package has been added
- exception -> error
- tweaks to accommodate for failing tests
- tests are complete for file/archive fetchers
- sniffer has been updated a bit to reflect kotlin changes for XHTML
- [Properties](https://github.com/readium/webpub-manifest/blob/master/properties.md) has been implemented as a generic map type with functions (JSON parse/unparse will be done later, during parser implementation)

pkg/fetcher/fetcher_archive.go
```diff
@@ -0,0 +1,150 @@
...
+	// Nothing needs to be done at the moment
+}
+
+func (r *entryResource) Link() pub.Link {
+	cl := r.entry.CompressedLength()
+	if cl == 0 {
+		cl = r.entry.Length()
+	}
+	r.link.Properties.Add(pub.Properties{
+		"archive": pub.Properties{
```
mickael-menu, line 112: I suggest using the `https://readium.org/webpub-manifest/properties#` prefix for internal properties. We can always add a section on this page to add some explanations for each internal property.

```suggestion
		"https://readium.org/webpub-manifest/properties#archive": pub.Properties{
```

pkg/mediatype/mediatype.go
```diff
@@ -0,0 +1,291 @@
...
+	Parameters map[string]string // The parameters in the media type, such as `charset=utf-8`.
+
+	// Private until used
+	// name          string
+	fileExtension string
+}
+
+// Create a new MediaType.
+// When an error is returned, do not use the resulting MediaType, as it will be incomplete/invalid
+func NewMediaType(str string, name string, extension string) (mt MediaType, err error) {
```
mickael-menu, line 36: Since we have a dedicated `mediatype` package, we can simplify a bit the name of the constructors (as well as the `MediaTypeOfX` sniffing functions).

```suggestion
func New(str string, name string, extension string) (mt MediaType, err error) {
```

From the coding convention:
> When a function in package `pkg` returns a value of type `pkg.Pkg` (or `*pkg.Pkg`), the function name can often omit the type name without confusion
> https://go.dev/blog/package-names

pkg/mediatype/mediatype_of.go
```diff
@@ -0,0 +1,142 @@
...
+		if err == nil {
+			return &mt
+		}
+	}
+
+	return nil
+}
+
+// Resolves a format from a list of mediatypes, list of extensions, and list of sniffers
+func MediaTypeOf(mediaTypes []string, extensions []string, sniffers []Sniffer) *MediaType {
```
mickael-menu, line 94:
```suggestion
func Of(mediaTypes []string, extensions []string, sniffers []Sniffer) *MediaType {
```

pkg/mediatype/mediatype_of.go
```diff
@@ -0,0 +1,142 @@
...
+	return mediaTypeOf(nil, mediaTypes, extensions, sniffers)
+}
+
+func MediaTypeOfStringAndExtension(mediaType string, extension string) *MediaType {
```
mickael-menu, line 98:
```suggestion
func OfStringAndExtension(mediaType string, extension string) *MediaType {
```

pkg/mediatype/mediatype_of.go
```diff
@@ -0,0 +1,142 @@
...
+	return mediaTypeOf(nil, []string{mediaType}, nil, Sniffers)
+}
+
+// Resolves a format from a single file extension
+func MediaTypeOfExtension(extension string) *MediaType {
+	return mediaTypeOf(nil, nil, []string{extension}, Sniffers)
+}
+
+// Resolves a format from a file
+func MediaTypeOfFile(file *os.File, mediaTypes []string, extensions []string, sniffers []Sniffer) *MediaType {
```
mickael-menu, line 113: In almost all cases we won't need to customize the `sniffers` when calling these functions. So let's simplify a bit and make it the default. We can add a `OfFileWithSniffers()`.

```suggestion
func OfFile(file *os.File, mediaTypes []string, extensions []string) *MediaType {
```

pkg/mediatype/types.go
```diff
@@ -0,0 +1,60 @@
+package mediatype
+
+var AAC, _ = NewMediaType("audio/aac", "", "aac")
```
mickael-menu, line 3: Maybe `const` instead of `var` here? Unless it would be useful to allow overwriting the default media types, I'm a bit on the fence for this one.

```suggestion
const AAC, _ = NewMediaType("audio/aac", "", "aac")
```

pkg/pub/contributor.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"encoding/json"
+	"errors"
+)
+
+// Contributor
+// https://github.com/readium/webpub-manifest/tree/master/contexts/default#contributors
+// https://github.com/readium/webpub-manifest/schema/contributor-object.schema.json
+type Contributor struct {
+	LocalizedName   LocalizedString  `json:"name" validate:"required"` // The name of the contributor.
```
mickael-menu, line 12:
```suggestion
	// The name of the contributor.
	LocalizedName   LocalizedString  `json:"name" validate:"required"`
```

pkg/pub/link.go
```diff
@@ -0,0 +1,57 @@
...
+	"path"
+
+	"github.com/readium/go-toolkit/pkg/mediatype"
+)
+
+// Link
+// https://github.com/readium/webpub-manifest/blob/master/README.md#24-the-link-object
+// https://github.com/readium/webpub-manifest/blob/master/schema/link.schema.json
+type Link struct {
+	Href       string     `json:"href"`                 // URI or URI template of the linked resource.
```
mickael-menu, line 13:
```suggestion
	// URI or URI template of the linked resource.
	Href       string     `json:"href"`
```

pkg/pub/manifest.go
```diff
@@ -0,0 +1,172 @@
...
+// Manifest Main structure for a publication
+type Manifest struct {
+	Context         []string `json:"@context,omitempty"`
+	Metadata        Metadata `json:"metadata"`
+	Links           []Link   `json:"links"`
+	ReadingOrder    []Link   `json:"readingOrder,omitempty"`
+	Resources       []Link   `json:"resources,omitempty"` //Replaces the manifest but less redundant
+	TableOfContents []Link   `json:"toc,omitempty"`
+
+	Subcollections map[string][]PublicationCollection `json:"-"` //Extension point for collections that shouldn't show up in the manifest
```
mickael-menu, line 17: What do you mean by shouldn't show up in the manifest?

For example the `landmarks` collection role from the EPUB profile is a useful subcollection to get for reading apps: https://github.com/readium/webpub-manifest/blob/master/profiles/epub.md

mickael-menu (changes requested):
@chocolatkey Could you check out the failing CI build as well? Thanks!

pkg/mediatype/types.go
chocolatkey: Unfortunately you can't make this const, it's simply not valid go because it's not a constant

pkg/archive/archive.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func NewArchiveFactory() *DefaultArchiveFactory {
+	return &DefaultArchiveFactory{}
+}
+
+// Holds an archive entry's metadata.
+type Entry interface {
+	Path() string                                // Absolute path to the entry in the archive.
```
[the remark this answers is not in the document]
chocolatkey: Comments to the right of struct/interface fields is actually standard Go, in all I've seen, and tools have no problem finding the comments.. Examples:
image

https://golang.org/src/go/doc/example.go
https://golang.org/doc/effective_go#formatting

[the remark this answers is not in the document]
mickael-menu: Sound good 👌

pkg/mediatype/types.go
mickael-menu: Alright, `var` it is then

pkg/pub/manifest.go
chocolatkey: Those are not my comments/exclusion behavior, they were kept from the previous code: https://github.com/readium/go-toolkit/blob/0bed5dd41b94da5eca1f904f6346877136134db3/models/publication.go#L27 . I have not yet touched manifest.go beyond the bare minimum (will be doing for parser).

mickael-menu: Okay, so for the next PR `Internal` can go as well as all the helper functions. Only these two are useful at this point:

https://github.com/readium/kotlin-toolkit/blob/b33a5dd5973aa4a896927df8878b82d03d42b8ee/readium/shared/src/main/java/org/readium/r2/shared/publication/Manifest.kt#L46-L58

Some of the helper functions will be added to the `Publication` struct.

chocolatkey:
@mickael-menu Do you see why https://github.com/readium/go-toolkit/runs/3760707837?check_suite_focus=true#step:4:185 is failing? `go build -v ./...` works on my computer

mickael-menu:
@chocolatkey The error was not super visible but it can't find `PublicationCollection` here:

https://github.com/readium/go-toolkit/blob/f7df767d42716f7f639d76a79c2d72ae762b3cf6/pkg/pub/manifest.go#L17

chocolatkey:
@mickael-menu thanks, I accidentally forgot to commit the file before because it was causing issues with the circular dependencies.

There is now a new issue with a failing test that I can't seem to replicate, both in Windows and WSL (Linux), where the mimetypes of some files is not as expected. The diff from the test is here: https://www.diffchecker.com/5n46ZVnt

mickael-menu:
Yes I can reproduce on macOS, my guess is that the standard `mime` package (`SniffSystem()`) behaves differently on each platform. You can probably work around this by adding an explicit sniffer to recognize the extensions `xml` as `application/xml` and `opf` as `application/oebps-package+xml`.

chocolatkey:
@mickael-menu thanks again, I added some code to initialize the go mimes with some that may not be available across OSes. We may need to add more later, but for now this passes the tests

mickael-menu (approved):
Great, pretty convenient we can add some explicitly in `mime`. That should be enough for now, thanks 👍

PR #61, opened 2022-03-16
MVP Position List implementation

chocolatkey:
- Add missing presentation properties to manifest
- Add new convenience function to publication for finding a resource in the manifest based on a (cleaned) path
- Add position list service
  - Needed to implement services concept
  - Locator-related structs and tests
  - EPUB position service, Per resource position service (for e.g. cbz)
  - No Audio position service - should it be included in an MVP? (not very difficult to implement)


pkg/manifest/locator.go
```diff
@@ -0,0 +1,190 @@
...
+import (
+	"encoding/json"
+
+	"github.com/pkg/errors"
+)
+
+// One or more alternative expressions of the location.
+// https://github.com/readium/architecture/tree/master/models/locators#the-location-object
+type Locations struct {
+	Fragments        []string               `json:"fragments,omitempty"`        // Contains one or more fragment in the resource referenced by the [Locator].
```
mickael-menu, line 12: Since you implemented `MarshalJSON()`, I guess the `json:` tags are redundant?

pkg/manifest/locator.go
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * There are many different use cases for locators:
+ *  - getting back to the last position in a publication
+ *  - bookmarks
+ *  - highlights & annotations
+ *  - search results
+ *  - human-readable (and shareable) reference in a publication
+ *
+ * https://github.com/readium/architecture/tree/master/locators
+ */
```
mickael-menu, line 140:
```suggestion
// Locator provides a precise location in a publication in a format that can be stored and shared.
//
// There are many different use cases for locators:
//  - getting back to the last position in a publication
//  - bookmarks
//  - highlights & annotations
//  - search results
//  - human-readable (and shareable) reference in a publication
//
// https://github.com/readium/architecture/tree/master/locators
```

pkg/manifest/locator.go
```diff
@@ -0,0 +1,190 @@
...
+type Locator struct {
+	Href      string     `json:"href"`
+	Type      string     `json:"type"`
+	Title     string     `json:"title,omitempty"`
+	Locations *Locations `json:"locations,omitempty"`
+	Text      *Text      `json:"text,omitempty"`
```
mickael-menu, line 146:
```suggestion
	Locations Locations `json:"locations,omitempty"`
	Text      Text      `json:"text,omitempty"`
```

pkg/manifest/locator.go
```diff
@@ -0,0 +1,190 @@
...
+// https://github.com/readium/architecture/tree/master/models/locators#the-location-object
+type Locations struct {
+	Fragments        []string               `json:"fragments,omitempty"`        // Contains one or more fragment in the resource referenced by the [Locator].
+	Progression      float64                `json:"progression,omitempty"`      // Progression in the resource expressed as a percentage (between 0 and 1).
+	Position         uint                   `json:"position,omitempty"`         // An index in the publication (>= 1).
+	TotalProgression float64                `json:"totalProgression,omitempty"` // Progression in the publication expressed as a percentage (between 0 and 1).
+	OtherLocations   map[string]interface{} `json:"otherLocations"`             // Additional locations for extensions.
+}
+
+func LocationsFromJSON(rawJson map[string]interface{}) (*Locations, error) {
```
mickael-menu, line 19: Using value semantics could be more convenient and safer for the `Locator` models.
```suggestion
func LocationsFromJSON(rawJson map[string]interface{}) (Locations, error) {
```

pkg/manifest/locator.go
```diff
@@ -0,0 +1,190 @@
...
+ */
+type Locator struct {
+	Href      string     `json:"href"`
+	Type      string     `json:"type"`
+	Title     string     `json:"title,omitempty"`
+	Locations *Locations `json:"locations,omitempty"`
+	Text      *Text      `json:"text,omitempty"`
+}
+
+func LocatorFromJSON(rawJson map[string]interface{}) (*Locator, error) {
```
mickael-menu, line 149:
```suggestion
func LocatorFromJSON(rawJson map[string]interface{}) (Locator, error) {
```

pkg/service/service.go
```diff
@@ -0,0 +1,117 @@
+package service
+
+import (
+	"github.com/readium/go-toolkit/pkg/fetcher"
+	"github.com/readium/go-toolkit/pkg/manifest"
+)
+
+// Base interface to be implemented by all publication services.
+type Service interface {
```
mickael-menu, line 9: Since services are core to the `Publication` object, we can keep these under the `pub` package. I think it makes for a clearer type too: `pub.PositionsService` vs `service.Positions`.

pkg/parser/epub/positions_service.go
```diff
@@ -1,5 +1,195 @@
...
+/**
+ * Positions Service for an EPUB from its [readingOrder] and [fetcher].
+ *
+ * The [presentation] is used to apply different calculation strategy if the resource has a
+ * reflowable or fixed layout.
+ *
+ * https://github.com/readium/architecture/blob/master/models/locators/best-practices/format.md#epub
+ * https://github.com/readium/architecture/issues/101
+ */
+type EPUBPositionsService struct {
```
mickael-menu, line 20: Since we're in the `epub` package and qualified types are mandatory in Go, I think we can simplify this into `epub.PositionsService`.

```suggestion
type PositionsService struct {
```

`epub.PositionsService` vs `epub.EPUBPositionsService`.

>The importer of a package will use the name to refer to its contents, so exported names in the package can use that fact to avoid repetition.
> https://go.dev/doc/effective_go#package-names

pkg/parser/epub/positions_service.go
```diff
@@ -1,5 +1,195 @@
...
+
+func (s *EPUBPositionsService) Links() []manifest.Link {
+	return []manifest.Link{service.PositionsLink}
 }
+
+func (s *EPUBPositionsService) Get(link manifest.Link) (fetcher.Resource, bool) {
+	return service.GetForPositionsService(s, link)
+}
+
+// Positions implements PositionsService
```
mickael-menu, line 38: Then, to disambiguate use the qualified name in the doc comments.

```suggestion
// Positions implements pub.PositionsService
```

pkg/parser/epub/positions_service.go
```diff
@@ -1,5 +1,195 @@
...
+
+/**
+ * Positions Service for an EPUB from its [readingOrder] and [fetcher].
+ *
+ * The [presentation] is used to apply different calculation strategy if the resource has a
+ * reflowable or fixed layout.
+ *
+ * https://github.com/readium/architecture/blob/master/models/locators/best-practices/format.md#epub
+ * https://github.com/readium/architecture/issues/101
+ */
```
mickael-menu, line 19:
```suggestion
// Positions Service for an EPUB from its [readingOrder] and [fetcher].
//
// The [presentation] is used to apply different calculation strategy if the resource has a
// reflowable or fixed layout.
//
// https://github.com/readium/architecture/blob/master/models/locators/best-practices/format.md#epub
// https://github.com/readium/architecture/issues/101
```

pkg/parser/epub/positions_service.go
```diff
@@ -1,5 +1,195 @@
...
+		}
+	}
+}
+
+/**
+ * Strategy used to calculate the number of positions in a reflowable resource.
+ *
+ * Note that a fixed-layout resource always has a single position.
+ */
+type ReflowableStrategy interface {
```
mickael-menu, line 147:
```suggestion
// Strategy used to calculate the number of positions in a reflowable resource.
//
// Note that a fixed-layout resource always has a single position.
type ReflowableStrategy interface {
```

pkg/parser/epub/positions_service.go
```diff
@@ -1,5 +1,195 @@
...
+
+	return uint(math.Max(math.Ceil(float64(length)/float64(l.PageLength)), 1))
+}
+
+/**
+ * Recommended historical strategy: archive entry length split by 1024 bytes pages.
+ *
+ * This strategy is used by Adobe RMSDK as well.
+ * See https://github.com/readium/architecture/issues/123
+ */
```
mickael-menu, line 194:
```suggestion
// Recommended historical strategy: archive entry length split by 1024 bytes pages.
//
// This strategy is used by Adobe RMSDK as well.
// See https://github.com/readium/architecture/issues/123
```

pkg/parser/parser_epub.go
```diff
@@ -8,10 +8,21 @@ import (
 	"github.com/readium/go-toolkit/pkg/mediatype"
 	"github.com/readium/go-toolkit/pkg/parser/epub"
 	"github.com/readium/go-toolkit/pkg/pub"
+	"github.com/readium/go-toolkit/pkg/service"
 	"github.com/readium/go-toolkit/pkg/util"
 )
 
 type EPUBParser struct {
```
mickael-menu, line 15: Since we now have an `epub` package, maybe this could be moved under `epub.Parser`?

pkg/pub/publication.go
```diff
@@ -11,40 +16,108 @@ type Publication struct {
...
 
-func New() *Publication {
-	return &Publication{}
+func New(m manifest.Manifest, f fetcher.Fetcher, b *service.ServicesBuilder) *Publication {
+	if b == nil {
+		b = service.NewBuilder(nil, nil, nil, nil, nil)
+	}
+	services := b.Build(service.NewContext(m, f)) // Build the services
+	var newManifest manifest.Manifest
+	copier.Copy(&newManifest, &m) // Make a deep copy of the manifest
```
mickael-menu, line 87: Why do we need to do a deep copy if the `Manifest` has value semantics? Are there nested reference properties?

pkg/service/service.go
```diff
@@ -0,0 +1,117 @@
...
+	serviceFactories map[string]ServiceFactory
+}
+
+func NewBuilder(
+	contentProtection ServiceFactory,
+	cover ServiceFactory,
+	locator ServiceFactory,
+	positions ServiceFactory,
+	search ServiceFactory,
+) *ServicesBuilder {
```
mickael-menu, line 42: I feel like this API, at least in Go, will be pretty error prone: easy to invert two parameters as there are of the same type and we will add more services from time to time which means duplicating this API to keep backward compatibility.

In Swift/Kotlin we can/must expand the parameter names which alleviates this issue.

Maybe we can just add a `NewBuilder()` that takes a single `map[string]ServiceFactory`?

pkg/service/service_positions.go
```diff
@@ -0,0 +1,92 @@
...
+func (s PerResourcePositionsService) Positions() []manifest.Locator {
+	poss := s.PositionsByReadingOrder()
+	positions := make([]manifest.Locator, len(poss))
+	for i, v := range poss {
+		positions[i] = v[0] // Always just one element
+	}
+	return positions
+}
+
+func (s PerResourcePositionsService) PositionsByReadingOrder() [][]manifest.Locator {
```
mickael-menu, line 64: We're missing helpers for these two APIs on the `Publication` object. Since we will move the positions service into the `pub` package, you should be able to add them easily.

mickael-menu (changes requested):
Thank you, I left a few comments. Nothing major except maybe renaming a couple packages.

pkg/manifest/locator.go
chocolatkey: Yes, that's true in many places in the toolkit. I left them just for humans to be able to easily tell what the JSON key of something is, but I can remove them if you'd like

mickael-menu: No that's actually a good idea, I was just a bit confused.

Could you just remove `json:"otherLocations"` as this one is incorrect? Thanks.

pkg/parser/parser_epub.go
chocolatkey: Done

pkg/manifest/locator.go
chocolatkey: Could you explain why you are suggesting this? According to https://github.com/readium/architecture/blob/master/models/locators/README.md#the-locator-object the locations and text properties are optional, and therefore the nicest JSON you can get would omit these properties if they have no value. If they aren't pointers, but direct values, then they will be the struct with the fields initialized to their default values. So, for example, a position object would go from having no `text` property, to having `text: {}`

pkg/manifest/locator.go
chocolatkey: This is to deal with the case where the raw JSON being inputted is nil. Would you prefer nil JSON become a default-initialized Locations object? That *may* result in some empty objects where they shouldn't be (or if unmarshalling and them marshalling again). Note this only applies because it's part of the unmarshalling/marshalling logic, normally I would agree values are safer

pkg/manifest/locator.go
chocolatkey: Same as above

pkg/pub/publication.go
chocolatkey: I'm replicating what was done here: https://github.com/readium/kotlin-toolkit/blob/540c80fc1c57f04d09db45764f103724177aedf9/readium/shared/src/main/java/org/readium/r2/shared/publication/Publication.kt#L88 which was to deep copy the manifest and add the service links

pkg/service/service.go
chocolatkey: Agreed, it looks much better

pkg/service/service_positions.go
chocolatkey: Can you elaborate on what this would look like? Are you saying a function that wraps `publication.Get(Link{Href: "/~readium/positions.json"})` ? Or more of a function that specifically finds the positions service, and specifically calls those functions?

pkg/manifest/locator.go
mickael-menu: The rational was:

* Safer as the locations and text can't be mutated from outside of the current scope on a `Locator` instance you hold.
* Easier to use when manipulating a `Locator`, because you don't need to check whether `Locations` or `Text` are `null` when accessing their properties. Instead you can do directly `locator.Locations.Progression`, or directly set `locator.Text.Before = "before"`. This will be particularly relevant when implementing APIs such as the search service.

Having empty JSON objects would be alright I guess. I'd prefer if we keep the models safe and not impacted by [the shortcoming of Go's serialization library](https://go-review.googlesource.com/c/go/+/13914/).

pkg/manifest/locator.go
mickael-menu: I think it depends on what we decide with https://github.com/readium/go-toolkit/pull/61#discussion_r918430208

pkg/pub/publication.go
mickael-menu: Kotlin doesn't have struct/value objects unfortunately, everything works by reference. So we had to make an actual copy of the manifest.

But if I'm not mistaken we don't need it with Go, even if the struct contains arrays. I did a quick test: https://go.dev/play/p/GfOWdK7sWMK

pkg/service/service_positions.go
mickael-menu: Sure, you can find it there: https://github.com/readium/kotlin-toolkit/blob/540c80fc1c57f04d09db45764f103724177aedf9/readium/shared/src/main/java/org/readium/r2/shared/publication/services/PositionsService.kt#L57-L85

I don't think we'll need the `positionsFromManifest` for the Go toolkit though, so it might just look like this:

```go
func (p *Publication) PositionsByReadingOrder() [][]manifest.Locator {
    service := p.FindService(PositionsService)
    if service == nil {
        return nil
    }
    return service.PositionsByReadingOrder()
}

func (p *Publication) Positions() [][]manifest.Locator {
    service := p.FindService(PositionsService)
    if service == nil {
        return nil
    }
    return service.Positions()
}
```

This way the caller can do this directly, without caring (or knowing about) the positions service:

```go
publication := Parse(epubFile)
publication.Positions()
```

mickael-menu (changes requested):
@chocolatkey Looking good besides the xmlquery dep 👍  Could you just update the title (and maybe description) of the PR? There's a lot more than just the positions service.

pkg/pub/publication.go
```diff
@@ -33,6 +31,41 @@ func (p Publication) JSONManifest() (string, error) {
 	return string(bin), nil
 }
 
+func (p Publication) PositionsFromManifest() []manifest.Locator {
+	// TODO just access the service directly and don't marshal and unmarshal JSON?
```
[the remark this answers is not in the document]
chocolatkey: Good point, I'll remove that comment

pkg/archive/archive_zip.go
```diff
@@ -145,3 +166,17 @@ func (e gozipArchiveFactory) OpenBytes(data []byte, password string) (Archive, e
...
 	return NewGoZIPArchive(r, func() error { return nil }), nil
 }
+
+type ReaderAtCloser interface {
+	io.Closer
+	io.ReaderAt
+}
+
+func (e gozipArchiveFactory) OpenReader(reader ReaderAtCloser, size int64, password string) (Archive, error) {
+	// Go's built-in zip reader doesn't support passwords. Maybe look into something like https://github.com/mholt/archiver
```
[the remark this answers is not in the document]
chocolatkey: I'd like to leave it in the interface for:
- Staying similar to the other toolkits
- Support for other formats, possibly DRMed
- Someone may want to write an implementation of ArchiveFactory, and the password option should be left there if they need it
I have altered gozipFactory to refuse to extract if a password is provided, to prevent confusion/misunderstanding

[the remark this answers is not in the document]
mickael-menu: Sounds good.

mickael-menu (approved):

PR #63, opened 2023-01-09
Add a basic `readium manifest` CLI utility

mickael-menu:
This PR adds a new `readium` command to expose command-line utilities from the toolkit.

Each utility will be its own subcommand. For now there's only a `readium manifest` command which prints out a RWPM.

I left the `server` command separate, but we could discuss about whether or not adding it as `readium serve`. 

## Manifest subcommand

```
Generate a Readium Web Publication Manifest for a publication.

This command will parse a publication file (such as EPUB, PDF, audiobook, etc.)
and build a Readium Web Publication Manifest for it. The JSON manifest is
printed to stdout.

Examples:
  Print out a minimal JSON RWPM.
  $ readium manifest publication.epub

  Pretty-print a JSON RWPM using tow-space indent.
  $ readium manifest --indent "  " publication.epub

  Extract the publication title with `jq`.
  $ readium manifest publication.epub | jq -r .metadata.title

Usage:
  readium manifest <pub-path> [flags]

Flags:
  -h, --help            help for manifest
  -i, --indent string   Indentation used to pretty-print
```


Makefile
```diff
@@ -0,0 +1,15 @@
+help:
```
chocolatkey, line 1: I'd suggest we use goreleaser (https://goreleaser.com) to build the `readium` binary, which we can also run in gh actions to generate new releases.
I guess this is OK as a temporary solution.

cmd/readium/cmd/root.go
```diff
@@ -0,0 +1,26 @@
...
+	Use:   "readium",
+	Short: "Readium Go toolkit",
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
```
chocolatkey, line 20: No need to log/print the error?

cmd/readium/cmd/manifest.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		return nil
+	},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		path := filepath.Clean(args[0])
+		pub, err := streamer.New(streamer.Config{}).Open(
+			asset.File(path), "",
+		)
+		if err != nil {
+			return fmt.Errorf("open failed: %w", err)
```
chocolatkey, line 50: I think this error message could be improved to contain the file path `fmt.Errorf("failed opening %s: %w")`

cmd/readium/cmd/manifest.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		var jsonBytes []byte
+		if indentFlag == "" {
+			jsonBytes, err = json.Marshal(pub.Manifest)
+		} else {
+			jsonBytes, err = json.MarshalIndent(pub.Manifest, "", indentFlag)
+		}
+		if err != nil {
+			return fmt.Errorf("json failed: %w", err)
```
chocolatkey, line 60: I think this error message could be improved to contain the file path `fmt.Errorf("failed rendering JSON for %s: %w")`.

chocolatkey (approved):
Looks good overall

cmd/readium/cmd/root.go
mickael-menu: Yes curious, it was generated by `cobra-cli`.

mickael-menu: Apparently the error is already printed by Cobra, this check is only to set the exit code.

Makefile
mickael-menu: It requires installing a third-party dependency to build the project locally though. However to generate GH release it definitely makes sense.

PR #64, opened 2023-01-09
Parse EPUB and RWPM accessibility metadata

mickael-menu:
Parse RWPM accessibility metadata introduced in https://github.com/readium/webpub-manifest/pull/85.


pkg/manifest/a11y.go
```diff
@@ -0,0 +1,407 @@
...
+	return A11y{
+		ConformsTo:            []A11yProfile{},
+		AccessModes:           []A11yAccessMode{},
+		AccessModesSufficient: [][]A11yPrimaryAccessMode{},
+		Features:              []A11yFeature{},
+		Hazards:               []A11yHazard{},
+	}
+}
+
+func (a A11y) IsEmpty() bool {
```
chocolatkey, line 34: This is a good approach, it might be a good idea to replicate this in the other structs in this repo

pkg/manifest/a11y.go
```diff
@@ -0,0 +1,407 @@
...
+
+	conformsTo, err := parseSliceOrString(rawJSON["conformsTo"], true)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed unmarshalling 'conformsTo'")
+	}
+	a.ConformsTo = A11yProfilesFromStrings(conformsTo)
+
+	if certJSON, ok := rawJSON["certification"].(map[string]interface{}); ok {
+		c := A11yCertification{
+			CertifiedBy: parseOptString(certJSON["certifiedBy"]),
```
chocolatkey, line 55: Just want to note (sort of as a reminder to myself) that eventually all the `parseOptXXX` functions should be replaced with a `parseOptional` that uses generics

pkg/manifest/a11y.go
```diff
@@ -0,0 +1,407 @@
...
+)
+
+func A11yHazardsFromStrings(strings []string) []A11yHazard {
+	return fromStrings(strings, func(str string) A11yHazard {
+		return A11yHazard(str)
+	})
+}
+
+func fromStrings[T any](strings []string, transform func(string) T) []T {
+	res := []T{}
```
chocolatkey, line 402: There is an optimization that can be made here to reduce allocations from growing the slice:
```go
res := make([]T, 0, len(strings))
```

pkg/parser/epub/metadata.go
```diff
@@ -378,21 +378,66 @@ func (m MetadataParser) computeMetaItem(expr MetadataItem, metas map[string][]Me
...
 	if !ok {
 		return ""
 	}
-	if len(p) == 0 {
-		return ""
+	return item.value
+}
+
+func (m metadataAdapter) Values(property string) []string {
+	values := []string{}
```
chocolatkey, line 406: Same optimization can be made in this function:
```go
var values []string
```
and then below the `ok`
```go
values = make([]string, 0, len(items))
for _, item := range items {
	values = append(values, item.value)
}
// or //
values = make([]string, len(items))
for i, item := range items {
	values[i] = item.value
}

File: pkg/parser/epub/metadata.go [hunk 5]
```diff
@@ -378,21 +378,66 @@ func (m MetadataParser) computeMetaItem(expr MetadataItem, metas map[string][]Me
...
+	if items, ok := m.items[property]; ok {
+		for _, item := range items {
+			values = append(values, item.value)
+		}
+	}
+	return values
+}
+
+func (m metadataAdapter) Links(rel string) []EPUBLink {
+	links := []EPUBLink{}
```
chocolatkey, line 416: For this function, because we can't predict the size very well, it's fine just to let it grow with append so I don't suggest the same optimization.

pkg/parser/epub/metadata.go
```diff
@@ -588,6 +634,149 @@ func (m PubMetadataAdapter) LocalizedSortAs() *manifest.LocalizedString {
...
+	}
+	return &c
+}
+
+func (m PubMetadataAdapter) a11ySummary() string {
+	return m.FirstValue(VocabularySchema + "accessibilitySummary")
+}
+
+func (m PubMetadataAdapter) a11yAccessModes() []manifest.A11yAccessMode {
+	am := []manifest.A11yAccessMode{}
```
chocolatkey, line 735: Same optimization for allocation can be made

pkg/parser/epub/metadata.go
```diff
@@ -588,6 +634,149 @@ func (m PubMetadataAdapter) LocalizedSortAs() *manifest.LocalizedString {
...
+	for _, v := range m.Values(VocabularySchema + "accessMode") {
+		am = append(am, manifest.A11yAccessMode(v))
+	}
+	return am
+}
+
+func (m PubMetadataAdapter) a11yAccessModesSufficient() [][]manifest.A11yPrimaryAccessMode {
+	ams := [][]manifest.A11yPrimaryAccessMode{}
```
chocolatkey, line 743: Same optimization for allocation can be made

pkg/parser/epub/metadata.go
```diff
@@ -588,6 +634,149 @@ func (m PubMetadataAdapter) LocalizedSortAs() *manifest.LocalizedString {
...
+		c := a11yAccessModesSufficient(v)
+		if len(c) > 0 {
+			ams = append(ams, c)
+		}
+	}
+	return ams
+}
+
+func a11yAccessModesSufficient(value string) []manifest.A11yPrimaryAccessMode {
+	ams := []manifest.A11yPrimaryAccessMode{}
```
chocolatkey, line 754: Same optimization for allocation can be made

pkg/parser/epub/metadata.go
```diff
@@ -588,6 +634,149 @@ func (m PubMetadataAdapter) LocalizedSortAs() *manifest.LocalizedString {
...
+		v = strings.TrimSpace(v)
+		if v != "" {
+			ams = append(ams, manifest.A11yPrimaryAccessMode(v))
+		}
+	}
+	return ams
+}
+
+func (m PubMetadataAdapter) a11yFeatures() []manifest.A11yFeature {
+	am := []manifest.A11yFeature{}
```
chocolatkey, line 765: Same optimization for allocation can be made

pkg/parser/epub/metadata.go
```diff
@@ -588,6 +634,149 @@ func (m PubMetadataAdapter) LocalizedSortAs() *manifest.LocalizedString {
...
+	for _, v := range m.Values(VocabularySchema + "accessibilityFeature") {
+		am = append(am, manifest.A11yFeature(v))
+	}
+	return am
+}
+
+func (m PubMetadataAdapter) a11yHazards() []manifest.A11yHazard {
+	am := []manifest.A11yHazard{}
```
chocolatkey, line 773: Same optimization for allocation can be made

chocolatkey (changes requested):
Looks good overall

pkg/manifest/a11y.go
mickael-menu: That would be nice, I have no experience with generics in Go so far.

pkg/manifest/a11y.go
mickael-menu: Makes sense, thanks.

pkg/parser/epub/metadata.go
mickael-menu: This one is not easy because we ignore empty lists, so we don't know in advance the length of the array.

pkg/parser/epub/metadata.go
mickael-menu: This one is not easy because we ignore blank values, so we don't know in advance the length of the array.

pkg/parser/epub/metadata.go
chocolatkey: Yes, but it doesn't really matter as long as we're close enough. It's better to over-allocate once rather than repeatedly allocate, and the amount we're over-allocating by is very small anyway. By doing `make(X, 0, len(Y))` we're just growing the capacity of the slice, but not the size, so if we end up overestimating the capacity, the length of the slice is still correct

pkg/parser/epub/metadata.go
chocolatkey: Same here

chocolatkey (changes requested):

pkg/manifest/a11y.go
chocolatkey: I'll do this after you merge. Also you say "no experience", but then I see `fromStrings` using generics ;)

mickael-menu: True, I guess I have one experience then. 😄 I think I adapted one of your generic functions for this one.

pkg/parser/epub/metadata.go
mickael-menu: Alright, let me update it.

chocolatkey (approved):
LGTM

PR #65, opened 2023-01-16
Infer `accessibility` and `numberOfPages` metadata

mickael-menu:
## `numberOfPages` inference

When the `metadata.numberOfPages` property is missing from the manifest, use the `--infer-page-count` flag to fill it using the generated position list.

## Accessibility metadata inference

`rwp manifest` can infer additional accessibility metadata when they are missing, with the `--infer-a11y` flag.

```sh
rwp manifest --infer-a11y publication.epub  | jq .metadata.accessibility
```

##### Inferred metadata

| Key | Value | Inferred? |
|-----|-------|-----------|
| `accessMode` | `auditory` | If the publication contains a reference to an audio or video resource (inspect `resources` and `readingOrder` in RWPM) |
| `accessMode` | `visual` | If the publications contains a reference to an image or a video resource (inspect `resources` and `readingOrder` in RWPM) |
| `accessModeSufficient` | `textual` | If the publication is partially or fully accessible (WCAG A or above)<br>Or if the publication does not contain any image, audio or video resource (inspect "resources" and "readingOrder" in RWPM)<br>Or if the only image available can be identified as a cover |
| `feature` | `displayTransformability` | If the publication is fully accessible (WCAG AA or above)<br>:warning: This property should only be inferred for reflowable EPUB files as it doesn't apply to other formats (FXL, PDF, audiobooks, CBZ/CBR). |
| `feature` | `printPageNumbers` | If the publications contains a page list (check for the presence of a `pageList` collection in RWPM) |
| `feature` | `tableOfContents` | If the publications contains a table of contents (check for the presence of a `toc` collection in RWPM) |
| `feature` | `MathML` | If the publication contains any resource with MathML (check for the presence of the `contains` property where the value is `mathml` in `readingOrder` or `resources` in RWPM) |
| `feature` | `synchronizedAudioText` | If the publication contains any reference to Media Overlays (TBD in RWPM) |

### DIFF
--- a/pkg/streamer/a11y_infer.go
+++ b/pkg/streamer/a11y_infer.go
@@ -1,0 +1,118 @@
+package streamer
+
+import (
+	"github.com/readium/go-toolkit/pkg/internal/extensions"
+	"github.com/readium/go-toolkit/pkg/manifest"
+	"github.com/readium/go-toolkit/pkg/mediatype"
+)
+
+func inferA11yMetadataFromManifest(mf manifest.Manifest) *manifest.A11y {
+	var a11y manifest.A11y
+	if mf.Metadata.Accessibility != nil {
+		a11y = *mf.Metadata.Accessibility
+	} else {
+		a11y = manifest.NewA11y()
+	}
+
+	accessModes := a11y.AccessModes
+	accessModesSufficient := a11y.AccessModesSufficient
+	features := a11y.Features
+
+	addFeature := func(f manifest.A11yFeature) {
+		if !extensions.Contains(features, f) {
+			features = append(features, f)
+		}
+	}
+
+	allResources := append(mf.ReadingOrder, mf.Resources...)
+
+	if len(accessModes) == 0 {
+		for _, link := range allResources {
+			if link.MediaType().IsAudio() || link.MediaType().IsVideo() {
+				accessModes = append(accessModes, manifest.A11yAccessModeAuditory)
+				break
+			}
+		}
+
+		for _, link := range allResources {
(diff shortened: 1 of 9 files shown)


pkg/streamer/a11y_infer.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		if !setTextual {
+			setTextual = true
+			for _, link := range allResources {
+				mt := link.MediaType()
+				if mt.IsAudio() ||
+					mt.IsVideo() ||
+					(mt.IsBitmap() && !extensions.Contains(link.Rels, "cover")) ||
+					mt.Matches(&mediatype.PDF) {
```
mickael-menu, line 64: @chocolatkey Do you remember why we pass `MediaType` pointers instead of values to `Matches()`? This doesn't feel right as media types are value-based.

chocolatkey: @mickael-menu That was quite a while ago and I've forgotten. But I'm going to guess it's because a nil value is an acceptable possibility to provide as a potential match (which will of course match false) and you don't have to nil check.

But then I thought a bit more, and remembered if something looks strange in this way it's probably because I was trying to stay close to what the kotlin toolkit did, and sure enough we have a [nullable MediaType in there](https://github.com/readium/kotlin-toolkit/blob/main/readium/shared/src/main/java/org/readium/r2/shared/util/mediatype/MediaType.kt#L196), so that's almost certainly why:
image
And I was/am replicating the exact logic in the function `Matches` wraps, the contains function which also lets you pass in null values it just ignores
image

chocolatkey (approved):
LGTM

PR #66, opened 2023-01-16
Add EPUB version metadata extension

mickael-menu:
Add a new metadata extension `http://www.idpf.org/2007/opf#version` to store the EPUB package version in the manifest.


chocolatkey (approved):
LGTM

PR #67, opened 2023-01-18
WIP PDF parsing

chocolatkey:
WIP PDF parser. Due to issues with pdfcpu (parsing of some PDFs causing panics, inefficient reading when streaming) it's not recommended for use in production. Plan is to fork pdfcpu to make fixes, then finish up/improve the parser

Other changes:
- Switch from []manifest.Link to manifest.LinkList where possible
- Bugfixes and improvements to fetchers' data reading functions
- Add fetcher.Resource io.ReadSeeker interface implementation
- Slice allocation efficiency improvements throughout code


pkg/parser/pdf/parser_metadata.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+
+	err = ParsePDFMetadata(ctx, &m)
+	if err != nil {
+		return
+	}
+
+	if m.Metadata.Identifier == "" {
+		// Create a UUIDv5 which is based on a SHA1 hash of the PDF's metadata.
```
mickael-menu, line 74: We used to generate a MD5 hash of the source file on the mobile toolkits (and maybe still do in some cases), but we're moving away from this and just use a `null` identifier if there's no intrinsic identifier in a publication format (or we can't get it).

For PDF, the specification actually mentions the `ID` array.

> A PDF file identifier is defined in the PDF specification as an array of two strings, the first of which is a permanent identifier that doesn’t change even when the file is updated. The second string changes each time the file is updated. For more information, see PDF Reference: Version 1.3 (Second Edition), Adobe Systems Incorporated.

Which I think you can get in `pdfcpu` with this: https://github.com/pdfcpu/pdfcpu/blob/07d97625e3fa00f05a6f7559a53463574329931a/pkg/pdfcpu/xreftable.go#LL101C3-L101C3

On the Swift toolkit, I used only the first identifier in the array (the stable one) and ignored the second one (updated every time the PDF is remade), and converted the bytes into an hexadecimal string:

https://github.com/readium/swift-toolkit/blob/e1735017b7c46ac62599e63adde961367a987b6f/Sources/Shared/Toolkit/PDF/CGPDF.swift#L20-L36

pkg/parser/pdf/parser_metadata.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// Keywords
+	if ctx.Keywords != "" && len(m.Metadata.Subjects) == 0 {
+		m.Metadata.Subjects = append(m.Metadata.Subjects, manifest.Subject{
+			LocalizedName: manifest.NewLocalizedStringFromString(ctx.Keywords),
+		})
+	}
+
+	// ctx.Producer
+	// ctx.Creator
```
mickael-menu, line 168:
```suggestion
```

pkg/parser/pdf/parser_metadata.go
chocolatkey: Thanks for the pointer to the PDF ID, I had forgotten about that because I was too focused on the xmp metadata. I've switched to using that first if it's there.
I definitely don't want to hash the file, because that's very inefficient in streaming use cases. From my testing, sometimes the identifier is already a hex string (like `88EC969638074476AC5DDC6638B9BB5E`) but can also be bytes in others, so I copied the hex conversion. Will also keep identifier a bit more consistent if syncing of data between toolkits is necessary.

mickael-menu (approved):

PR #68, opened 2023-01-20
Use the default Meta vocabulary for unqualified metadata

mickael-menu:
When an EPUB (2 or 3) used an EPUB 2 meta property, it was added to the `OtherMetadata` map without a vocabulary, which was then expanded in the RWPM.

```xml
<meta content="1.9.20" name="Sigil version"/>
```

```json
"metadata": {
    "title": "Alice in wonderland",
    "Sigil version": "1.9.20"
}
```

To clean that up, I prefixed unqualified properties with the Meta Properties vocabulary.

> The [Meta Properties Vocabulary](https://www.w3.org/TR/epub/#app-meta-property-vocab) is the [default vocabulary](https://www.w3.org/TR/epub/#sec-default-vocab) for use with the property attribute.
> https://www.w3.org/TR/epub/#sec-meta-elem

```json
"metadata": {
    "title": "Alice in wonderland",
    "http://idpf.org/epub/vocab/package/meta/#Sigil version": "1.9.20"
}
```
It's not perfect as it's supposed to be for EPUB 3, but that gives a way for users to query unqualified metadata no matter what the source EPUB version is.

### DIFF
--- a/pkg/parser/epub/metadata.go
+++ b/pkg/parser/epub/metadata.go
@@ -196,7 +196,7 @@ func (m MetadataParser) parseMetaElement(element *xmlquery.Node) *MetadataItem {
 		if content == "" {
 			return nil
 		}
-		resolvedName := resolveProperty(name, m.prefixMap, NoVocab)
+		resolvedName := resolveProperty(name, m.prefixMap, DefaultVocabMeta)
 		return &MetadataItem{
 			property: resolvedName,
 			value:    content,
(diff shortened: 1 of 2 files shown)


chocolatkey:
@mickael-menu shouldn't we be escaping the property name in that example you gave? `http://idpf.org/epub/vocab/package/meta/#Sigil version` is technically not a valid URL/URI because of that space.

mickael-menu:
Yes good idea, I'll make the changes.

chocolatkey (approved):
LGTM

PR #69, opened 2023-02-07
Auto-release the CLI tools with GoReleaser

mickael-menu:
This PR adds a GoReleaser config and GitHub action to automatically build artifacts and create a new GitHub release when pushing a new tag.

Note that this doesn't sign and notarize the binaries, so macOS will refuse to run them unless you use a right click, then "Open" with Shift held.

https://github.com/mitchellh/gon could be a lead for supporting notarization.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: Release
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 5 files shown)


chocolatkey (approved):
LGTM

PR #75, opened 2023-03-03
Fix computing the positions list of an FXL EPUB

mickael-menu:
There was an issue when computing the EPUB positions list if a reading order item didn't have the `layout` link property. Instead of using the publication `metadata.presentation.layout` property, the fallback was `reflowable`.


chocolatkey (approved):
LGTM

PR #76, opened 2023-03-10
Split or merge inferred accessibility metadata

mickael-menu:
#### Accessibility inference

`rwp manifest` can infer additional accessibility metadata when they are missing, with the `--infer-a11y` flag. It takes one of the following arguments:

| Option           | Description                                                                                            |
|------------------|--------------------------------------------------------------------------------------------------------|
| `no` (*default*) | No accessibility metadata will be inferred.                                                            |
| `merged`         | Accessibility metadata will be inferred and merged with the authored ones in `metadata.accessibility`. |
| `split`          | Accessibility metadata will be inferred but stored separately in `metadata.inferredAccessibility`.     |

```sh
rwp manifest --infer-a11y=merged publication.epub  | jq .metadata
```

### DIFF
--- a/pkg/manifest/metadata.go
+++ b/pkg/manifest/metadata.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/readium/go-toolkit/pkg/internal/util"
 )
 
 // TODO replace with generic
@@ -127,19 +128,29 @@ func (m Metadata) InferredAccessibility() *A11y {
 // SetOtherMetadata marshalls the value to a JSON map before storing it in
 // OtherMetadata under the given key.
 func (m Metadata) SetOtherMetadata(key string, value interface{}) error {
-	bytes, err := json.Marshal(value)
+	value, err := toJSONMap(value)
 	if err != nil {
 		return err
+	}
+	m.OtherMetadata[key] = value
+	return nil
+}
+
+func toJSONMap(value interface{}) (map[string]interface{}, error) {
+	if value, ok := value.(util.JSONMappable); ok {
+		return value.JSONMap()
+	}
+
+	bytes, err := json.Marshal(value)
+	if err != nil {
+		return nil, err
 	}
 	var object map[string]interface{}
 	err = json.Unmarshal(bytes, &object)
 	if err != nil {
-		return err
+		return nil, err
 	}
-
-	m.OtherMetadata[key] = object
-
-	return nil
+	return object, nil
 }
 
 func MetadataFromJSON(rawJson map[string]interface{}, normalizeHref LinkHrefNormalizer) (*Metadata, error) {
--- a/pkg/manifest/metadata.go
+++ b/pkg/manifest/metadata.go
@@ -112,6 +112,34 @@ func (m Metadata) EffectiveReadingProgression() ReadingProgression {
 	return LTR
 }
 
+// InferredAccessibility returns the accessibility metadata inferred from the
+// manifest and stored in OtherMetadata.
+func (m Metadata) InferredAccessibility() *A11y {
+	var a11y *A11y
+	if a11yJSON, ok := m.OtherMetadata["inferredAccessibility"].(map[string]interface{}); ok {
+		a11y, _ = A11yFromJSON(a11yJSON)
+	}
+	return a11y
+}
+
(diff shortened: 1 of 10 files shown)


HadrienGardeur:
Yeah, let's use a URI instead, it's more consistent with our approach for extensibility.

pkg/manifest/metadata.go
```diff
@@ -112,6 +112,34 @@ func (m Metadata) EffectiveReadingProgression() ReadingProgression {
...
+	var object map[string]interface{}
+	err = json.Unmarshal(bytes, &object)
+	if err != nil {
+		return err
+	}
+
+	m.OtherMetadata[key] = object
+
+	return nil
+}
```
chocolatkey, line 141: Can you explain why you're marhsalling and then unmarshalling the passed-in value? Is this in order to do a deep copy?

pkg/streamer/streamer.go
```diff
@@ -32,12 +32,25 @@ type Streamer struct {
 type Config struct {
 	Parsers              []parser.PublicationParser // Parsers used to open a publication, in addition to the default parsers.
 	IgnoreDefaultParsers bool                       // When true, only parsers provided in parsers will be used.
-	InferA11yMetadata    bool                       // When true, additional accessibility metadata will be infered from the manifest.
+	InferA11yMetadata    InferA11yMetadata          // When set, additional accessibility metadata will be infered from the manifest.
```
chocolatkey, line 35: I'd say "not empty" instead of "set", since you can't "unset" this by e.g. nulling it. But it's good enough either way

pkg/streamer/streamer.go
```diff
@@ -32,12 +32,25 @@ type Streamer struct {
...
 	InferPageCount       bool                       // When true, will infer `Metadata.NumberOfPages` from the generated position list.
 	ArchiveFactory       archive.ArchiveFactory     // Opens an archive (e.g. ZIP, RAR), optionally protected by credentials.
 	HttpClient           *http.Client               // Service performing HTTP requests.
 }
 
+type InferA11yMetadata string
```
chocolatkey, line 41: Why a string instead of e.g. uint8 with values 1,2,3 (0 is unset)? It would be more efficient, and since there are const values defining the options, it's not very different for implementers. The only different is needing to translate from command line string to the option

chocolatkey (changes requested):
Overall, good

pkg/streamer/streamer.go
mickael-menu: Good idea. I started from the CLI side and moved this to the Streamer without much thought. The translation from strings is already done anyways.

pkg/manifest/metadata.go
mickael-menu: `OtherMetadata` must contain "raw" JSON objects (as in `map[string]interface{}`) and not complex types. Because we want the same results as if this was parsed from a `manifest.json` file.

I didn't find a way to go directly from `struct` -> `map[string]interface{}`, hence the weird dance with `Marshall`/`Unmarshall`.

chocolatkey: @mickael-menu how about this? https://github.com/mitchellh/mapstructure . If you want I can build a performance benchmark to compare, but my gut instinct tells me this will definitely be more performant than the JSON way.

mickael-menu: Wouldn't this bypass any custom `Marshall` implementation in our structs?

mickael-menu: Which API would you use for this? I see only decoding APIs, but here we want to encode a native struct to `map[string]interface{}` using the JSON rules of each struct.

chocolatkey: Yeah, on second thought you're right... However, I have a new idea: We add an interface like this:
```go
interface Mappable {
    ToMap() (map[string]interface{}, error)
}
```
And the function first attempts to load the supplied `value` as something that implements that interface and call that function. If that's not possible, then it falls back to your json approach.

mickael-menu: Sounds good 👍

mickael-menu: I named it `JSONMappable` to make it explicit that the objects in the map must be JSON-compatible.

chocolatkey (approved):
Maybe change uint to uint8, otherwise LGTM

PR #79, opened 2023-05-16
Add new a11y metadata inference rules

mickael-menu:
* Infer `visual` `accessModeSufficient` for publications containing only image or video resources.
* Infer `textual` `accessMode` with publications conforming to WCAG A, or referencing no media resources.
* Infer `synchronizedAudioText` feature when a publication references SMIL resources.

### DIFF
--- a/pkg/manifest/link.go
+++ b/pkg/manifest/link.go
@@ -296,6 +296,18 @@ func (ll LinkList) AllAreVideo() bool {
 	return true
 }
 
+// Returns whether all the resources in the collection are bitmaps or video clips.
+func (ll LinkList) AllAreVisual() bool {
+	for _, link := range ll {
+		mt := link.MediaType()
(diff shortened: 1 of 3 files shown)


pkg/manifest/link.go
```diff
@@ -296,6 +296,18 @@ func (ll LinkList) AllAreVideo() bool {
...
 }
 
+// Returns whether all the resources in the collection are bitmaps or video clips.
+func (ll LinkList) AllAreVisual() bool {
+	for _, link := range ll {
+		mt := link.MediaType()
+		if !mt.IsBitmap() && !mt.IsVideo() {
+			return false
+		}
+		// TODO should we check alternates?
```
mickael-menu, line 306: I kept this TODO as it is a good question. We don't support alternates at the moment so `false` is reasonable, but that might change.

PR #81, opened 2023-09-05
Fix `textual` access mode inference with FXL

mickael-menu:
* We can't really infer a `textual` access mode with a FXL EPUB, even if it doesn't contain any image, because the DOM itself might be garbled (e.g. absolute positioning of letters/words). The inference was updated to only check reflowable EPUBs.
* EPUB manifest items that are fallbacks of a spine item are now also added to the `resources` list in the RWPM. Often, these resources are actually embedded in the main XHTML resource. Not having them in `resources` produced false positives in the a11y inference as the bitmaps in `Link.children` were not taken into account.


chocolatkey (approved):
LGTM

PR #82, opened 2023-09-07
Content iterator

chocolatkey:
Need to port over the tests too before this is ready

Notable changes:
- Bump minimum Go version from 1.18 to 1.21
- Adds content iterator based on the mobile toolkits, and also implemented in the ts-toolkit
- Improvements to minimal reads mode for ZIPs
- Change `manifest.Properties` to avoid mutations, also adding a `Properties()` getter to the `Resource` interface. See https://github.com/readium/go-toolkit/discussions/88 for more info
- Add deduplication for Contributors, a common issue in EPUB metadata
- Fix concurrent map access fault in marshalling of manifest
- Improve EPUB font deobfuscator
- Fix bug in XML query of META-INF/display-options.xml

Future improvements:
- Add the memory-efficient content iterator improvements as with the ts-toolkit
- Full test suite for content iterator


pkg/pub/service_content.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (s DefaultContentService) Content(start *manifest.Locator) content.Content {
+	return ContentImplementation{
+		context:                          s.context,
+		start:                            start,
+		resourceContentIteratorFactories: s.resourceContentIteratorFactories,
+	}
+}
+
+type ContentImplementation struct {
```
mickael-menu, line 71: I think the implementation can stay private here?

```suggestion
type contentImplementation struct {
```

pkg/util/css.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	return ""
+}
+
+// Get a CSS selector that will uniquely select a provided HTML element
+// If the element has an ID, returns #id;
+// otherwise returns the parent (if any) CSS selector, followed by '>'
+// followed by a unique selector for the element (tag.class.class:nth-child(n)).
+// Logic copied from JSoup: https://github.com/jhy/jsoup/blob/0b10d516ed8f907f8fb4acb9a0806137a8988d45/src/main/java/org/jsoup/nodes/Element.java#L829
+func CSSSelector(n *html.Node) string {
```
mickael-menu, line 60: This will become part of the public API surface. Shouldn't this be under `pkg/internal/util`?

pkg/content/content.go
```diff
@@ -0,0 +1,53 @@
...
+	"strings"
+
+	"github.com/readium/go-toolkit/pkg/content/element"
+	"github.com/readium/go-toolkit/pkg/content/iterator"
+)
+
+type Content interface {
+	Text(separator *string) (string, error) // Extracts the full raw text, or returns null if no text content can be found.
+	Iterator() iterator.Iterator            // Creates a new iterator for this content.
+	Elements() ([]element.Element, error)   // Returns all the elements as a list.
```
mickael-menu, line 13: In the Kotlin toolkit, `text()` and `elements()` are in the interface with a default implementation, but they should probably be extensions instead. The only thing that should be required to implement is `Iterator`.

Maybe we can remove them from the interface and keep only `ContentText()` and `ContentElements()` on the side?

Alternatively if we want to keep the API easy to use, `Content` could be a struct and have a `IteratorFactory` interface instead.

pkg/content/element/element.go
```diff
@@ -0,0 +1,260 @@
...
+// [Iterator], where it's the return type of many of the functions. Maybe we should
+// rethink this approach with all the interfaces later when not copying the kotlin.
+
+// Represents a single semantic content element part of a publication.
+type Element interface {
+	// AttributesHolder
+	Language() string
+	AccessibilityLabel() string
+	Attribute(key AttributeKey) *Attribute[any]
+	Attributes(key AttributeKey) []Attribute[any]
```
mickael-menu, line 21: How about just `Attributes() AttributesHolder` to simplify the implementations?

pkg/content/iterator/html.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/readium/go-toolkit/pkg/manifest"
+	"github.com/readium/go-toolkit/pkg/mediatype"
+	"golang.org/x/net/html"
+	"golang.org/x/net/html/atom"
+)
+
+type HTMLContentIterator struct {
+	resource        fetcher.Resource
+	locator         manifest.Locator
+	BeforeMaxLength int // Locators will contain a `before` context of up to this amount of characters.
```
mickael-menu, line 19: This property doesn't need to be public, but we should be able to set it when calling `NewHTML()` or using `HTMLFactory`.

```suggestion
	beforeMaxLength int // Locators will contain a `before` context of up to this amount of characters.
```

pkg/content/iterator/iterator.go
```diff
@@ -0,0 +1,73 @@
+package iterator
+
+import "github.com/readium/go-toolkit/pkg/content/element"
+
+// Iterates through a list of [Element] items asynchronously.
+// [hasNext] and [hasPrevious] refer to the last element computed by a previous call to any of both methods.
+// TODO: It's based on a kotlin iterator, maybe we can make this more of something for go?
```
mickael-menu, line 7: I didn't find much related to Go iterator, it looks like it's not very idiomatic. 🤷

pkg/manifest/locator.go
```diff
@@ -114,6 +116,19 @@ func (l Locations) MarshalJSON() ([]byte, error) {
...
+func (l Locations) CSSSelector() string {
+	if v, ok := l.OtherLocations["cssSelector"]; ok {
+		if s, ok := v.(string); ok {
+			return s
+		}
+	}
+	return ""
+}
+
+// TODO partialCfi and domRange getters
```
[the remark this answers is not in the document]
danielweck: Thorium doesn't really use / need EPUB CFI. We generate them in our "locators" next to the data structure we actually rely on to make things work internally (DOM Range for precise locations, percentage progression otherwise ... oh, and by the way we still don't support positions list yet). We emit CFI "just because we can", and it's only the path to the element anyway, no character-level reference. To achieve feature-parity with our internal DOM Range (de)serialisation mechanism, we would need a properly-tested and conformant CFI implementation, which apparently doesn't exist :)

* https://github.com/mbret/prose-reader/blob/master/packages/core/src/cfi/CfiHandler.ts
* https://github.com/fread-ink/epub-cfi-resolver
* https://github.com/vivliostyle/vivliostyle.js/blob/master/packages/core/src/vivliostyle/cfi.ts
* https://github.com/johnfactotum/foliate-js/blob/main/epubcfi.js
* https://github.com/futurepress/epub.js/blob/master/src/epubcfi.js -- https://github.com/futurepress/EpubCFI/blob/master/src/epubcfi.js
* https://github.com/readium/readium-cfi-js
* https://github.com/satorumurmur/bibi/blob/master/__src/bibi/extensions/epubcfi.js
* https://github.com/hypothesis/client/blob/main/src/shared/cfi.ts
* Colibrio? https://api-docs.colibrio.com/web/3.12.0/modules/colibrio_core_selector_epubcfi.html

Also see https://github.com/futurepress/epub.js/issues/1358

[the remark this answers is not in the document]
chocolatkey: This work is based on the kotlin-toolkit, which is presumably based on the documentation here: https://github.com/readium/architecture/blob/master/models/locators/extensions/html.md . `partialCfi` is just a string in kotlin, so it could theoretically omitted since you can access it if it exists via `OtherLocations`, but since `DomRange` is [implemented in the kotlin-toolkit](https://github.com/readium/kotlin-toolkit/blob/develop/readium/shared/src/main/java/org/readium/r2/shared/publication/html/DomRange.kt) I would want to implement it in the go-toolkit as well

pkg/mediatype/sniffer.go
```diff
@@ -143,6 +143,36 @@ func SniffBitmap(context SnifferContext) *MediaType {
...
+		return &OPUS
+	}
+	if context.HasFileExtension("wav") {
+		return &WAV
+	}
+	if context.HasFileExtension("webm") {
+		return &WEBMAudio
+	}
+
+	// TODO read magic bytes?
```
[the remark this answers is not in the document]
chocolatkey: @mickael-menu I've added the mimetype sniffing for all those audio formats that were missing.
For what you're suggesting though, I'm not sure how it would work. The sniffing functions are always returning specific mimetypes, not a generic audio, video etc. type. What we would return if the `audio/` prefix were detected?

pkg/parser/epub/parser.go
```diff
@@ -160,7 +164,7 @@ func parseDisplayOptions(fetcher fetcher.Fetcher) (ret map[string]string) {
 		}
 	}
 
-	if platform := displayOptionsXml.SelectElement("platform"); platform != nil {
+	if platform := displayOptionsXml.SelectElement("//platform"); platform != nil {
```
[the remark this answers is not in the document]
chocolatkey: @mickael-menu I don't think this is an issue with the mobile toolkits. The `kotlin-toolkit` version of this [uses a `getFirst` function](https://github.com/readium/kotlin-toolkit/blob/ac6392ccbd133698d83fc8b76c0fb3744e7197d1/readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/EpubParser.kt#L192) which is, as I understand, equivalent to what this change does. There was a commercial EPUB that was reported by a user as not showing as fixed layout in the web reader using the go-toolkit as a backend, so I investigated and it turns out this was an "Apple-optimized" FXL EPUB that only gave itself away as FXL in the `display-options.xml`, not in the OPF. That made me realize the query was faulty.

Side note, as I was looking into this, I found that there are some upstream changes that can be pulled into our copy of xmlquery: https://github.com/readium/xmlquery/pull/1/commits . The xmlpath library it uses [added a useful function to supply namespaces](https://github.com/antchfx/xpath/pull/82), so I might make an improvement soon that will improve on the somewhat excessive [XPATH query lengths](https://github.com/readium/go-toolkit/blob/content-iterator/pkg/parser/epub/utils.go#L32) as a result of namespaces

pkg/pub/publication.go
```diff
@@ -89,7 +113,9 @@ func (p Publication) Find(path string) *manifest.Link {
 		}
 	}
 
-	link.Href = "/" + link.Href
+	if !strings.HasPrefix(link.Href, "/") {
+		link.Href = "/" + link.Href
+	}
```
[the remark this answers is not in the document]
chocolatkey: @mickael-menu I would be more than happy to get rid of the `/` prefix. I'm honestly not sure why it was ever added, but I copied it for consistency's sake when porting from the kotlin-toolkit and it has been a source of irritation. I think it is also confusing for users of this toolkit, as there ends up being a lot of code adding and removing slashes, even within your own code. The relative paths are good and work. I'll work on this in another PR

pkg/pub/service_content.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/readium/go-toolkit/pkg/content/element"
+	"github.com/readium/go-toolkit/pkg/content/iterator"
+	"github.com/readium/go-toolkit/pkg/fetcher"
+	"github.com/readium/go-toolkit/pkg/manifest"
+)
+
+// TODO content iterator special ~readium link
+
+var ContentLink = manifest.Link{
+	Href: "/~readium/content.json",
```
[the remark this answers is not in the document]
chocolatkey: @mickael-menu I will address this in a separate PR along with the rest of the absolute path removals

pkg/util/css.go
chocolatkey: @mickael-menu you're correct. I've moved it

pkg/content/element/element.go
chocolatkey: @mickael-menu You mean make a `type AttributesHolder []Attribute[any]`?

pkg/content/iterator/html.go
chocolatkey: @mickael-menu Unfortunately because of the language constraints, we can't have an optional parameter to a function with a default value like what is done in the kotlin-toolkit. I think making the property public is acceptable because I imagine a situation like this, which is also used in other go libraries:
```go
// direct usage
x := iterator.NewHTML(resource, locator)
x.BeforeMaxLength = 123
x.HasNext()

// or based on the current ContentElements() func
it := content.Iterator()
it.(*iterator.HTMLContentIterator).BeforeMaxLength = 123
hasNext, err := it.HasNext()
```
This is the "substitute" for that language feature. I personally think requiring the beforeMaxLength to be specified in the initial function is "ugly", because it is something that will probably not be changed in typical usage, which is why the kotlin-toolkit makes it optional.

pkg/content/iterator/iterator.go
chocolatkey: Yeah, I'm keeping this in sync with kotlin for now so I'm not going to change it as part of this PR at least

pkg/manifest/locator.go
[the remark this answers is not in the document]
mickael-menu: The extension helpers are implemented but never used. It might not be worth implementing it in the Go toolkit, unless you do want to generate partial CFI or DOM ranges at some point.

pkg/mediatype/sniffer.go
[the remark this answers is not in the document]
mickael-menu: Ha you're right, I read too fast and thought this was analog to `MediaType.isAudio` in the Kotlin toolkit.

pkg/parser/epub/parser.go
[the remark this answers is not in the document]
mickael-menu: Thanks for the explanation 👌

pkg/pub/publication.go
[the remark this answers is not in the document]
mickael-menu: Sounds good, I created an issue to keep track of this: https://github.com/readium/go-toolkit/issues/90

pkg/content/element/element.go
mickael-menu: There's one already here: https://github.com/readium/go-toolkit/blob/ba558c981288f460e690a1450b424a0b2ed8e1db/pkg/content/element/attributes.go#L22-L24

pkg/content/iterator/html.go
mickael-menu: That's true, and I don't think we can customize it on the Kotlin toolkit anyways, unless you create an `HtmlResourceContentIterator` manually.

mickael-menu (approved):
LGTM, thank you @chocolatkey

PR #93, opened 2024-06-04
Compressed resource optimization

### LINKED ISSUE #89: Add ability to read raw compressed archive entries
When streaming an EPUB's content to web browsers using the go-toolkit, it is possible for the end user's device to be the only one that must perform decompression of the file's contents. This would result in a further decrease in CPU usage of any software using the go-toolkit, as decompression would not have to occur within the code, and additional reductions in resource usage may be avoided due to no longer needing to *recompress* the resource further up the chain, in e.g. nginx, apache etc.
How does this work in practice? Here's an example of the accept-encoding header of a modern browser (Chrome): `Accept-Encoding:
gzip, deflate, br, zstd`. The [`deflate`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding#deflate) encoding is the same compression scheme used inside of ZIP files. This means that if you were to provide a browser that supports this encoding the contents of a compressed `zip.File` using the `OpenRaw()` method, and return `content-encoding: deflate`, the browser would be able to decode it directly.

chocolatkey:
This enhancement allow for resources to be read in their compressed form, which usually applies to most if not all resources in any ZIP-based format (primarily EPUB). This saves CPU time because any modern web browser is capable of decompressing a `deflate`-encoded resource, so the raw contents of the ZIP can be directly proxied to any web client. Closes #89


HadrienGardeur:
This is related to #89

pkg/archive/archive.go
```diff
@@ -55,9 +55,10 @@ type Entry interface {
 	Path() string                                              // Absolute path to the entry in the archive.
 	Length() uint64                                            // Uncompressed data length.
 	CompressedLength() uint64                                  // Compressed data length.
+	CompressedAs(compressionMethod uint16) bool                // Whether the entry is compressed using the given method.
```
mickael-menu: Isn't the `8` value for deflate only relevant in the context of a ZIP? Theoretically we could use `Archive` for a CBR/RAR file as well.

To remove any ambiguity, what do you think of adding an explicit enum with `Deflate` case in the toolkit instead of using ints?

Having `CompressionMethod() CompressionMethod` might be more convenient too.

mickael-menu (commented):
Nice optimization 👍

pkg/archive/archive.go
chocolatkey: Technically, the latest version of the [ZIP spec](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) supports other compression formats as well. I think the correct term would be that the EPUB standard only supports modes 0 and 8. I think it's worth keeping the door open for other compression modes for other container formats that we might support, or whose specs may change to support other methods in the future.

I've added your suggested `CompressionMethod` enum

mickael-menu (approved):
Thank you for making the change 👌

PR #99, opened 2024-08-05
Remove the example webserver

chocolatkey:
The server example located at `cmd/server` has been made redundant by the `rwp serve` command, and improvements to the server have been made only in the `rwp` server codebase.
Furthermore, I don't think it's within the scope of this project's repository to provide a full-fledged server that makes decisions regarding the logging SDK to use, the configuration file format, the use of Sentry for error reporting, etc.
The `rwp serve` server is much easier to start and setup, and is therefore much more suitable for development and testing of the project, which is the primary intention of the web server.


chocolatkey:
I'm not sure how, but I accidentally merged this into the `develop` branch with a single commits sync button push in VS Code... Sorry...

mickael-menu (commented):
I never complain when we remove code anyway 😄

PR #129, opened 2024-10-03
Add safety check for invalid font obfuscation key

chocolatkey:
Some broken publications have an invalid font obfuscation key, and this causes a panic due to a divide by zero in the `deobfuscateFont` function. This adds a check for that


mickael-menu (approved):
LGTM 🙏

PR #139, opened 2024-10-09
Add mobile toolkit v3 URL helper

### LINKED ISSUE #90: Remove leading slashes from relative HREFs
To align with the [new v3 HREF strategy on the mobile toolkits](https://github.com/readium/go-toolkit/pull/82#discussion_r1607812022), leading slashes should not be added in relative HREFs.

See this comment: https://github.com/readium/go-toolkit/pull/82#discussion_r1609489862

chocolatkey:
This ports the kotlin-toolkit's V3 update's URL helper classes, to replace the current ones, which resolves #90

Other changes made include:
- Replace the mobile toolkits' custom implementation of URI templates with a more complete third-party implementation, which helps with service link matching
- Add the `ManifestTransformer` functionality from the mobile toolkit
- Moved to using native mediatypes instead of mime strings across the codebase, and added a few necessary helper functions
- Fix OPUS mimetype
- Implement the missing Locator `Text.Substring` function
- Added URL schemes, with support for cloud protocols, in preparation for cloud streaming


pkg/manifest/locator.go
```diff
@@ -160,26 +188,39 @@ func TextFromJSON(rawJson map[string]interface{}) (t Text) {
 //
 // https://github.com/readium/architecture/tree/master/locators
 type Locator struct {
-	Href      string    `json:"href"`
-	Type      string    `json:"type"`
-	Title     string    `json:"title,omitempty"`
-	Locations Locations `json:"locations,omitempty"`
-	Text      Text      `json:"text,omitempty"`
+	Href      url.URL              `json:"href"`
+	MediaType *mediatype.MediaType `json:"type"`
```
mickael-menu: Media types are required for locators, so I would not use a pointer here to enforce it.

```suggestion
	MediaType mediatype.MediaType `json:"type"`
```

pkg/parser/epub/parser_navdoc.go
```diff
@@ -88,21 +88,21 @@ func parseLiElement(li *xmlquery.Node, filePath string) (link *manifest.Link) {
 		title = strings.TrimSpace(muchSpaceSuchWowMatcher.ReplaceAllString(first.InnerText(), " "))
 	}
 	rawHref := first.SelectAttr("href")
-	href := "#"
+	href := url.MustURLFromString("")
```
mickael-menu: I don't think an empty string is a valid URL, does it pass the validation? We used a `#` when we needed a "null" URL.

pkg/util/url/url.go
```diff
@@ -0,0 +1,434 @@
...
+
+	"github.com/readium/go-toolkit/pkg/internal/extensions"
+	"golang.org/x/net/idna"
+)
+
+/*
+A Uniform Resource Locator.
+
+https://url.spec.whatwg.org/
+*/
```
mickael-menu:
```suggestion
// A Uniform Resource Locator.
//
// https://url.spec.whatwg.org/
```

pkg/util/url/url.go
```diff
@@ -0,0 +1,434 @@
...
+// Create a [URL] from a Go net/url URL.
+func URLFromGo(url gurl.URL) (URL, error) {
+	if url.IsAbs() {
+		return AbsoluteURLFromGo(url)
+	} else {
+		return RelativeURLFromGo(url)
+	}
+}
+
+// Represents an absolute Uniform Resource Locator.
```
mickael-menu:
```suggestion
// Represents a relative Uniform Resource Locator.
```

pkg/util/url/url.go
```diff
@@ -0,0 +1,434 @@
...
+/*
+A Uniform Resource Locator.
+
+https://url.spec.whatwg.org/
+*/
+type URL interface {
+	Path() string            // Decoded path segments identifying a location.
+	Filename() string        // Decoded filename portion of the URL path.
+	Extension() string       // Extension of the filename portion of the URL path.
+	Query() gurl.Values      // Returns the query parameters present in this URL.
```
mickael-menu: I would remove this API in favor of using straight `url.Raw().Query()` where needed for now. This way it is more future-proof if we need to add our own data structure representing query parameters, or if we switch to a different implementation of URL than `net/url`.

```suggestion
```

mickael-menu (commented):
Great work, that looks much cleaner! 👍

pkg/parser/epub/parser_navdoc.go
chocolatkey: @mickael-menu yes, it parses with Golang's URL parser:
image
Do you want me to change it back to an anchor link?

mickael-menu: You decide what you think is best. IMHO it would be more consistent as `#` is used elsewhere in the Go toolkit.

mickael-menu (approved):
Thanks for making the changes 🙏

PR #183, opened 2025-01-31
printPageNumbers --> pageNavigation

### LINKED ISSUE #92: Return `pageNavigation` instead of `printPageNumbers` when inferring metadata
`printPageNumbers` has been gradually deprecated in favor of two different values:

- [`pageBreakMarkers`](https://www.w3.org/community/reports/a11y-discov-vocab/CG-FINAL-vocabulary-20230718/#pageBreakMarkers) which is meant to indicate that the text of the publication contains page break markers using ARIA
- and [`pageNavigation`](https://www.w3.org/community/reports/a11y-discov-vocab/CG-FINAL-vocabulary-20230718/#pageNavigation) which is meant to indicate that the publication contains a list of pages, usually based on HTML IDs

The current inference technique is based on the presence of `pageList` in the RWPM output which can either come from:

- a Navigation Document in EPUB 3.x
- or an NCX in EPUB 2.x

Given the nature of what is inferred here, the toolkit should return `pageNavigation` instead of `printPageNumbers`.

chocolatkey:
Closes #92


HadrienGardeur (approved):

mickael-menu (approved):
Good to merge 👌

PR #195, opened 2025-02-16
Add EPUB a11y 1.1 Conformance Support

### LINKED ISSUE #194: Support for EPUB Accessibility 1.1 conformance values
In a recent PR on the RWPM repo (https://github.com/readium/webpub-manifest/pull/109), we introduced additional values for accessibility conformance that cover all the values currently supported in EPUB Accessibility 1.1.

To stay up to date with that work, we need to update the current parser to support all of these new values.

As a complementary feature, I'll also open an issue in the `ts-toolkit` to provide a helper that converts these URLs into a human-readable string.

chocolatkey:
Also implements sorting of conformsTo values from best to worst
Closes #194

### DIFF
--- a/pkg/manifest/a11y.go
+++ b/pkg/manifest/a11y.go
@@ -2,6 +2,7 @@ package manifest
 
 import (
 	"encoding/json"
+	"slices"
 
 	"github.com/pkg/errors"
 	"github.com/readium/go-toolkit/pkg/internal/extensions"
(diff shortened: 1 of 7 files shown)


HadrienGardeur (approved):

pkg/manifest/a11y.go
```diff
@@ -171,14 +174,70 @@ const (
...
+	}
+
+	pRank := a11yProfileRanking[p]
+	oRank := a11yProfileRanking[other]
+	return oRank - pRank
+}
+
+type A11yProfileList []A11yProfile
+
+func (l A11yProfileList) Sort() {
```
mickael-menu, line 232: Is this API really useful for integrators, or should we keep it internal for now? It seems we pre-sort the list during parsing.

chocolatkey: @mickael-menu I opted to keep it public because if users of the toolkit are crafting their own WebPub manifest from scratch (or from another source), they can use this function to take advantage of the sorting logic we already implemented to sort the array before e.g. marshalling or performing other logic, since it's only currently performed upon parsing of EPUB or WebPub. Do you think that's too obscure?

HadrienGardeur: Makes sense to me, I've heard about a few organizations that would like to distribute content that they author as HTML rather than EPUB, so they could be the target for that.

PR #211, opened 2025-04-01
Implement remote streaming of publications

chocolatkey:
- Both S3 (AWS and S3-compatible services) as well as Google Cloud Storage are supported
- Tests will definitely be failing right now. Want to see feedback first to avoid rewriting tests
- I'm still experimenting with the `fs` wrapper

Let me know what you think of the API. Unfortunately I did not have the time to also update to the Readium Mobile v3 AssetRetriever/PublicationOpener so it still uses the Streamer API.

The `rwp serve` server distinguishes between the location of the publication based on the URL scheme of the encoded path component. E.g. `http://wps-server/Z3M6Ly9yZWFkaXVtLXBsYXlncm91bmQtZmlsZXMvZGVtby9sZXNfZGlhYm9saXF1ZXMuZXB1Yg/OEBPS/content-7.xhtml` 's path decoded is `gs://readium-playground-files/demo/les_diaboliques.epub`. The acceptable URL schemes are `gs`, `s3`, `file`, `http`, `https`. For backwards-compatibility, "naked" paths (no URL scheme) will turn into file URLs.


pkg/archive/archive_s3.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// CanOpen implements SchemeSpecificArchiveFactory
+func (e S3ArchiveFactory) CanOpen(scheme url.Scheme) bool {
+	return scheme == url.SchemeS3
+}
+
+// Open implements ArchiveFactory
+func (e S3ArchiveFactory) Open(location url.URL, password string) (Archive, error) {
+	return nil, errors.New("S3 archives must be opened with OpenWithContext")
```
mickael-menu: I'm not sure what the `Context` is used for, but that does not feel right to have implementations implementing selectively some of the APIs of an interface.

Would it be possible to keep only the `Open()` API taking a `Context` for all factories? Or pass the `Context` object when instantiating the factory struct instead of passing it to `Open()`?

pkg/archive/remote.go
```diff
@@ -0,0 +1,469 @@
...
+		zipTailSize:         65 * 1024, // 65KB
+	}
+	if !r.cacheAll() {
+		r.cachedRanges = make([]readRange, 0, r.cacheCountThreshold)
+	}
+	return r
+}
+
+// S3-specific reader
+type remoteS3Reader struct {
```
mickael-menu: I only quickly brushed over this, but would it make sense to move the S3 and GC specific readers in the `archive/archive_s3.go` and `archive/archive_gc.go` files instead?

pkg/fetcher/fetcher.go
```diff
@@ -28,6 +32,30 @@ type Fetcher interface {
...
+
+	/**
+	 * Known resources available in the medium, such as file paths on the file system
+	 * or entries in a ZIP archive. This list is not exhaustive, and additional
+	 * unknown resources might be reachable.
+	 *
+	 * If the medium has an inherent resource order, it should be followed.
+	 * Otherwise, HREFs are sorted alphabetically.
+	 */
+	LinksWithContext(ctx context.Context) (manifest.LinkList, error)
```
mickael-menu: It looks like another symptom of the issue I talked about in the other comment. It would be more convenient if we can get the context from the properties of the implementing struct, or keep the context-based APIs for all cases.

pkg/fetcher/traits.go
```diff
@@ -14,3 +16,34 @@ type CompressedResource interface {
...
+	ReadAsXML(prefixes map[string]string) (*xmlquery.Node, *ResourceError)
+}
+
+// Adds context to resource retrieval.
+type RemoteResource interface {
+	Resource
+
+	// Returns data length from metadata if available, or calculated from reading the bytes otherwise.
+	// This value must be treated as a hint, as it might not reflect the actual bytes length. To get the real length, you need to read the whole resource.
+	LengthWithContext(ctx context.Context) (int64, *ResourceError)
```
mickael-menu: Same

pkg/fetcher/traits.go
```diff
@@ -14,3 +16,34 @@ type CompressedResource interface {
 	ReadCompressed() ([]byte, *ResourceError)
 	ReadCompressedGzip() ([]byte, *ResourceError)
 }
+
+type StringResource interface {
```
mickael-menu: Could you explain why you extracted these APIs in a new `StringResource` interface?

If I'm not mistaken, as we don't know if a `Resource` will contain plain text until we try to read it, you will need to implement this for all `Resource` structs?

As I can see there's a lot of code duplication for these implementations, I think these APIs don't fit Go very well. Maybe keeping only the top-level `ReadResourceAsString()` functions is enough? Alternatively I found this technique using a base embedded struct: https://go.dev/play/p/ByoUzfqOpn

pkg/fetcher/resource.go
```diff
@@ -243,6 +247,55 @@ func Other(cause error) *ResourceError {
...
+			if ae.ErrorCode() == "AccessDenied" {
+				return Forbidden(err)
+			}
+		}
+	}
+
+	return Other(err)
+}
+
+func GCSErrorToException(err error) *ResourceError {
```
mickael-menu: These don't look like they belong in this module?

pkg/archive/archive_s3.go
chocolatkey: The context is used to give the implementer an opportunity to cancel the requests that are being made to perform the initial opening of the archive, or have them time out, if there is no longer a reason to proceed with the opening of the archive. In the example below, you can see my reasoning for having it be in the function instead of the archive factory creation functions, namely that I don't think the factories should take in the context because otherwise they are no longer reusable across multiple requests in this example. Because a factory is (as far as I understand based on the mobile toolkits) supposed to be for create e.g. an archive over and over again, it wouldn't make sense to provide it with the context for only a single opening of the archive. Let me know if that makes sense.

```go
var archiveFactory = archive.NewGCSArchiveFactory(gcsClient, archive.NewDefaultRemoteArchiveConfig())

func handleRequest(r *http.Request) {
    archive, err := archiveFactory.OpenWithContext(r.Context()) // This will be cancelled if the HTTP request is cancelled because it uses the requests's context
    // ...
}
```

What I could theoretically do is what I did with some other structs, and keep the `Open` API functional as a fallback, using `context.Background()`. It isn't ideal to use though, and I'd like to discourage its usage because having infinite contexts could lead to a request to open an archive being stuck due to a network issue, and I'd rather have implementers not fall into that trap like I have before.

pkg/archive/remote.go
chocolatkey: Sure, I can move them

pkg/fetcher/fetcher.go
chocolatkey: Same issue as above. What do you mean by `keep the context-based APIs for all cases` though? Like, getting rid of the non-context APIs?

pkg/fetcher/traits.go
chocolatkey: I've never liked the duplication either. I'll look into that technique, maybe I can get rid of those duplicate functions. If not, I'll just turn it into utility functions instead if you don't mind it not matching the mobile toolkit

pkg/fetcher/resource.go
chocolatkey: I put them here because it's the best place I could think of, since it's not like we have a special package for remote/cloud streaming. I was thinking about creating `pkg/cloud`, but it seemed to bare-bones. What do you think would be the best place to place these?

pkg/fetcher/traits.go
mickael-menu: > If not, I'll just turn it into utility functions instead if you don't mind it not matching the mobile toolkit

Absolutely, I'm all in favor of keeping the APIs more idiomatic to the platform than exactly in sync between the toolkits.

pkg/fetcher/resource.go
mickael-menu: I would put `AWSErrorToException` in `fetcher_s3.go` and `GCSErrorToException` in `fetcher_gcs.go` where they are used, and probably keep them private.

pkg/archive/archive_s3.go
mickael-menu: Thank you, that makes perfect sense. It looks like this is a pretty generic Go way of cancelling a task, so I think it could make sense to get rid of the APIs that don't use `Context` and require it every time.

pkg/fetcher/fetcher.go
mickael-menu: Yes exactly.

pkg/fetcher/traits.go
chocolatkey: I've removed that interface and the repeated functions, the places where they're needed just use plain functions that perform the equivalent work

pkg/fetcher/resource.go
chocolatkey: Done

chocolatkey:
@mickael-menu I've moved off of the `*WithContext` functions, and refactored the codebase to work with `context.Context`. Any other issues/suggestions before I start with tests?

mickael-menu:
@chocolatkey Not that I can think of 🤔
 
Thank you for making these changes.

PR #217, opened 2025-04-17
Image inference

### LINKED ISSUE #176: Extracting additional information from images
We'd like to extract the following information from every bitmap image present in an EPUB package:

- [x] height
- [x] width
- [x] size of the file
- [x] presence of animated content
- [x] cryptographic hash
- [x] perceptual hash

This will be based on https://pkg.go.dev/golang.org/x/image plus additional packages for hashing.

All of these info will be added to the shared model for the Link Object and also included in the JSON/RWPM output:

- Link Objects can already document a `height` and a `width`
- We might consider adding the size of a file as a core property of the Link Object (`size`?)
- While the presence of animated content and crypto/perspectual hashes should rely on `properties` and extensions

chocolatkey:
Closes #176

- Includes the streaming working of #211 (no changes, no need to re-review)
- Add a helper to transform fetchers and resources into interfaces compatible with Go's `fs.FS` and `fs.File`
- Add support for `hash` property in links, with a list of recognized algorithms and utility functions
- Add `--inspect-images` and `--hash` flags to `rwp manifest` utility to enable inference of additional image metadata


cmd/rwp/cmd/manifest.go
```diff
@@ -20,6 +26,14 @@ var inferA11yFlag InferA11yMetadata
 // Infer the number of pages from the generated position list.
 var inferPageCountFlag bool
 
+/*var inferIgnoreImageHashesFlag []string
+
+var inferIgnoreImageDirectoryFlag string*/
```
mickael-menu, line 30: Is there a plan to restore these flags at some point?

cmd/rwp/cmd/manifest.go
```diff
@@ -121,3 +164,45 @@ func (e *InferA11yMetadata) Set(v string) error {
 func (e *InferA11yMetadata) Type() string {
 	return "string"
 }
+
+type ImageInspector struct {
```
mickael-menu: This file should only parse CLI arguments and then route the requested operations.

To avoid ending up with a massive file of unrelated stuff, I would move the `ImageInspector` to its own package or at least its own file in `cmd`.

pkg/analyzer/image.go
```diff
@@ -0,0 +1,382 @@
...
+	"github.com/readium/go-toolkit/pkg/mediatype"
+	"go4.org/media/heif"
+	"golang.org/x/exp/slices"
+	"golang.org/x/image/riff"
+	"golang.org/x/image/webp"
+)
+
+const blurHashAlgorithm = "https://blurha.sh"
+
+type ImageProperties struct {
```
mickael-menu: Could you add doc comments for all the new public APIs in this PR?

pkg/fetcher/resource.go
```diff
@@ -434,12 +396,222 @@ func (r ProxyResource) ReadCompressedGzip() ([]byte, *ResourceError) {
...
+func (r *LazyResource) ReadCompressedGzip(ctx context.Context) ([]byte, *ResourceError) {
+	cres, ok := r.resource().(CompressedResource)
+	if !ok {
+		return nil, Other(errors.New("resource is not compressed"))
+	}
+	return cres.ReadCompressedGzip(ctx)
+}
 
-// TODO BufferingResource
+// TODO FallbackResource, SynchronizedResource, BufferingResource
```
mickael-menu, line 617: Many of these resource implementations were added to fix issues on the other toolkits. I wouldn't bother porting them unless you face a problem that could be fixed with it.

pkg/manifest/properties_hash.go
```diff
@@ -0,0 +1,64 @@
...
+
+import "github.com/pkg/errors"
+
+type HashAlgorithm string
+
+// The following hashes keys are reserved for future use, but not necessarily supported by the toolkit.
+// If you are using a hash algorithm not listed here, it's better to use a URI, such as `https://blurha.sh`.
+// If there's an algorithm you think should be recognized, let us know.
+const (
+	HashAlgorithmBlake2b  HashAlgorithm = "blake2b"
```
mickael-menu, line 11: Isn't blur.sh supported by the toolkit? Why is it private and not part of this enum?

cmd/rwp/cmd/manifest.go
chocolatkey: @mickael-menu yes, those will be implemented once this code has been moved to the `readium-cli`

pkg/manifest/properties_hash.go
chocolatkey: @mickael-menu While it's supported by the toolkit, I didn't want to "immortalize" it and make it officially supported until @HadrienGardeur had a chance to think about this. His initial request was to support hashing of the file (thus the well-known hashing algorithms in that enum), as well as a perceptual hash, for which I chose `phash-dct`. The addition of blurhash was a bonus you requested, but I'm not sure if we should turn it into a "hash algorithm" since that's not really what it is, because the value is an actual thumbnail-sized representation of the image which cannot really be mathematically compared to another hash for equivalency or similarity. Let me know if you think differently

pkg/fetcher/resource.go
chocolatkey: OK, I thought they were added for the sake of users of the toolkit. I'll check through them and see what's worth implementing or not in the future, I've created this issue for it: https://github.com/readium/go-toolkit/issues/218

cmd/rwp/cmd/manifest.go
chocolatkey: @mickael-menu I've moved the excess code to a `helpers` package. Once we move this all to `readium-cli`, I'll use the `pkg` folder like with this repo

pkg/manifest/properties_hash.go
mickael-menu: I think your argument makes sense. Feel free to resolve this comment.

cmd/rwp/cmd/manifest.go
mickael-menu: Sounds good, thank you.

pkg/analyzer/image.go
chocolatkey: @mickael-menu Thanks for reminding me, that's done. I also made `ImageProperties` private, that was public due to a previous version of the `image.Image` function which returned it.

chocolatkey:
@mickael-menu regarding the dependency issue, as per the meeting I'm going to deal with this after the merge, as the perceptual hashing package (which is not maintained) itself depends on that image resizing dependency. Furthermore, we may want to move the image inspection code out of the go-toolkit into the readium-cli

mickael-menu (approved):
Thank you for making the changes

PR #228, opened 2025-05-05
Support ignoring images based on hash when inferring nonvisual reading

chocolatkey:
Toolkit portion of the work for readium/cli#15
From the changelog:

### Added

- New config option available when creating a `Streamer`: `InferIgnoredImages`, a list of hashes of images to ignore when when inferring nonvisual reading
- `analyzer.MatchImage` function that compares an image link's hashes with given hashes to check for a match
- `HashValue` has new `String` and `Equal` convenience functions. `HashList` has a new `Find` convenience function.

### Changed

- Renamed `analyzer.Image` to `analyzer.InspectImage`
- Slight adjustments to behavior of manifest properties functions


pkg/analyzer/image.go
```diff
@@ -343,21 +360,21 @@ func Image(system fs.FS, link manifest.Link, algorithms []manifest.HashAlgorithm
 	// TODO: rewrite more cleanly
 	s2hash := sha256.New()
 	mdhash := md5.New()
-	if slices.Contains(algorithms, manifest.HashAlgorithmSHA256) && slices.Contains(algorithms, manifest.HashAlgorithmMD5) {
+	if slices.Contains(neededAlgorithms, manifest.HashAlgorithmSHA256) && slices.Contains(algorithms, manifest.HashAlgorithmMD5) {
```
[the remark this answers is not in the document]
mickael-menu: @chocolatkey I think this one might be relevant?

pkg/analyzer/image.go
```diff
@@ -387,3 +404,51 @@ func isWEBPAnimated(file io.Reader) (bool, error) {
...
+	}
+	for _, hash := range hashes {
+		if v, ok := linkHashes.Find(hash.Algorithm); ok {
+			if v.Equal(hash) {
+				// Simple equality
+				return true, nil
+			}
+
+			// Special distance-based matching for perceptual hashes
+			if v.Algorithm == manifest.HashAlgorithmPhashDCT {
```
mickael-menu, line 427: This is perfectly fine for this PR. One possible improvement to support extensibility of hash algorithms would be to delegate the comparison to a set of `HashMatcher`, with the default one doing a basic equality for example.

pkg/manifest/properties.go
```diff
@@ -102,11 +102,18 @@ func (p Properties) Layout() EPUBLayout {
 }
 
 func (p Properties) Encryption() *Encryption {
-	mp, ok := p.Get("encrypted").(map[string]interface{})
+	v := p.Get("encrypted")
+	if v == nil {
+		return nil
+	}
+	if enc, ok := v.(Encryption); ok {
+		// Is already an Encryption struct
```
mickael-menu: Link `Properties` must only contain JSON data specifically to avoid this ambiguity and facilitate the serialization of the manifest to JSON. If there are `Encryption` or `HashList` objects in `Properties`, they need to be serialized at the location where they are added to `Properties`.

pkg/analyzer/image.go
chocolatkey: This might be something to add as an issue. I definitely agree if this gets any more complicated we'll need that

pkg/analyzer/image.go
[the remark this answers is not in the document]
chocolatkey: Yep, fixed

pkg/manifest/properties.go
chocolatkey: @mickael-menu I changed it, look OK?

mickael-menu (approved):
Looking good, thank you for making the changes 👌

PR #245, opened 2025-07-18
Remove presentation, add layout

### LINKED ISSUE #243: Migrate away from `presentation` in `metadata`
The `go-toolkit` is still using a `presentation` object in `metadata`: https://github.com/readium/go-toolkit/blob/develop/pkg/manifest/presentation.go

This object is based on a mix of EPUB and advanced Divina properties that have been mostly deprecated by now.

We'd like to move away from this approach and use the `layout` property directly in `metadata` instead with three values allowed: `fixed`, `reflowable` and `scrolled`. This property should be nullable and null by default.

For properties that can be found in EPUB (`orientation`, `flow` and `spread`), we should continue to parse them but they should show up using our extensibility rules with a full URL instead, expanded from their namespace.

chocolatkey:
Resolves #243, also:
- Update `metadata.readingProgression`
- Add `metadata.altIdentifier` & `contributor.altIdentifier`


pkg/streamer/a11y_infer.go
```diff
@@ -52,8 +52,7 @@ func inferA11yMetadataInPublicationManifest(ctx context.Context, pub *pub.Public
 	// if the only image available can be identified as a cover.
 	if !isTextual &&
 		mf.ConformsTo(manifest.ProfileEPUB) &&
-		mf.Metadata.Presentation != nil &&
-		*mf.Metadata.Presentation.Layout == manifest.EPUBLayoutReflowable {
+		mf.Metadata.EffectiveLayout() == manifest.LayoutReflowable {
```
chocolatkey, line 55: @HadrienGardeur is the condition that the publication must be EPUB-conforming still valid?

HadrienGardeur: I think that's still true:

- a valid EPUB must declare all resources contained in its package
- for Web Publications that do not conform to EPUB, this is less obvious currently, which means that they might omit resources that would make them partially inaccessible
- we can't expect text to be in the proper reading order in a fixed layout EPUB, which is why we must restrict this inference to reflowable EPUB publications
- for audiobooks, PDF and Divina, this inference doesn't apply

chocolatkey:
@HadrienGardeur As I am implementing `altIdentifier` in this PR, I am running into some questions:

I'm implementing the `altIdentifier` parsing right now, is there anything special I should do for EPUBs aside from take the string and add it to the `altIdentifier` array? I ask because it has the option to be a JSON object with value and scheme.
Looking at the EPUB spec it seems I probably need to do something. If you have some examples of e.g. an EPUB with lots of identifiers with schemes that would be helpful. It looks like there's `opf:scheme`, but also `identifier-type`, and possibly others like figuring out the scheme from the `urn:` prefix if it exists. Although that also applies to the primary `identifier`, where we can't attach a schema because it's just a string unlike the altIdentifiers (do we want to put the primary identifier a second time in the `altIdentifier` array?)

And regarding the `altIdentifier` for contributors, how do we go about that? Currently we just take the first identifier and it becomes the `identifier` for a contributor. I don't think there's a "primary identifier" specified in EPUB, so should we just have the first identifier be the primary and the rest alts?

chocolatkey:
What I've done for now is:
metadata identifier = EPUB `unique-identifier`
metadata altIdentifier = everything else, just the values
contributor identifier = first identifier
contributor altIdentifier = the rest of the identifiers (if any)

HadrienGardeur:
`altIdentifier` was primarily introduced for OPDS where we're more likely to encounter such metadata.

What you've implemented is good enough for EPUB, I don't think it's worth doing more than that.

mickael-menu (approved):
LGTM 👍

I believe the comments from Copilot are valid, you could consider having a non-optional `Layout` in the service.

I would also keep the `Correct` and `MinimalValue` APIs as private to the toolkit as well.

None of these issues are blocking to me though.

chocolatkey:
@mickael-menu thanks. The co-pilot comment is slightly incorrect about the compilation error (there is none) but yeah, it doesn't need to be a pointer anymore. The correct and minimalValue funcs are private now

PR #270, opened 2025-10-10
Services improvements

### LINKED ISSUE #62: --infer-page-count does not work
With the current (latest) version, using `--infer-page-count` flag on the `manifest` does not work.

In previous versions, it would extract a number of pages, which was visible under `metadata.numberOfPages`. That key is no loger returned.

Command used: `manifest --infer-a11y=split --infer-page-count path/to/file.epub`.

chocolatkey:
This issue's goal is to give implementers more control over the access of a service publicly (e.g. webserver) versus in Go code, and provide the necessary building blocks to fix https://github.com/readium/cli/issues/62 

# Changelog

All services are now hidden by default. This mainly affects implementers creating webservers, but you also shouldn't need to manually remove services just to produce clean manifest output. To restore previous functionality, set the `streamer.Config`'s `AddServiceLinks` to `true`

### Added

- ServicesBuilder now has a convenience function `Services` to get the names of all services currently in the builder
- ServicesBuilder now has a `ExposeLinks` and `HideLinks` function to toggle the exposure of a service via the links that get added to the WebPub manifest, as well as access to the service via its well-known link path. By default, services are **private**
- `streamer.Config` has a new property, `AddServiceLinks`. Setting this property is equivalent to calling the aforementioned `ExposeLinks` function for every service

### Changed

- `ServiceFactory` now has a required `public` property. This lets a service expose itself via the `Get` and `Links` function if set to true. This also means that by default, all services are now "private". That means they will not be added to manifests as links, or callable by said link's path. They are still directly accessible in Go code using e.g. `Publication.FindService`, and then directly calling their functions by casting them to the correct service type (see e.g. `Publication.Positions`)


mickael-menu (commented):
The implementation and solution are looking good 👍 

Just a doubt regarding the meaning of "public" / "publicize". Technically the services are always public in the Publication object (albeit with native Go APIs). What about using `ExposeAsResources` or `ExposeLinks` instead?

chocolatkey:
@mickael-menu I've changed it to `ExposeLinks` and `HideLinks`

mickael-menu (approved):
Thank you for making the changes.

PR #273, opened 2025-10-14
Fixes for inefficient reads of ZIP entries

### LINKED ISSUE #27: Server struggles with remote streaming of compressed video/audio
This isn't _quite_ a bug, but could be in some cases considered as such - the server included with the cli struggles to stream video and audio files that are compressed inside of a ZIP-based container, such as EPUB. This was encountered for example with one of the IDPF sample publications: https://github.com/IDPF/epub3-samples/releases/download/20230704/cc-shared-culture.epub
This publication contains compressed videos:
Image
When an asset is compressed using DEFLATE, it becomes impossible to seek to a specific position in the file without first decompressing all content before that offset (in practice, basically the entire file due to ZIP APIs in Golang). Modern browsers' native video players attempt to use byte ranges to buffer a portion of the video (or audio), so it can start playing and save data for the client. However because the video file is compressed, every time a small range is read, the entire video file must be decompressed over and over again. And because the video file is large, it is not cached by our remote streaming caching logic, which currently defaults to 1MB as the max file size for cache eligibility. This leads to a very inefficient and resource-heavy process on the server-side, and a slow loading for the user:
Image

Furthermore, the streaming of this large video seems to be triggered an additional bug in the server, where it is returning the wrong size of content, as logged by browsers: `net::ERR_CONTENT_LENGTH_MISMATCH 206 (Partial Content)`.
And the publication server we have running on Google Cloud Run returns a 502 gateway timeout: https://publication-server.readium.org/aHR0cHM6Ly9naXRodWIuY29tL0lEUEYvZXB1YjMtc2FtcGxlcy9yZWxlYXNlcy9kb3dubG9hZC8yMDIzMDcwNC9jYy1zaGFyZWQtY3VsdHVyZS5lcHVi/EPUB/video/shared-culture.mp4

This is also definitely something we should include in our optimization of EPUBs, as compressing of any media content (image, video, audio) make streaming worse

chocolatkey:
Fixes https://github.com/readium/cli/issues/27
- Fixes bugged `Read` calls that aren't obligated to return the entirety of a resource
- Add [zran](https://github.com/madler/zlib/blob/master/examples/zran.h)-like reading of ZIP entries. This only applies to the start offset, not the total length of the resources. Particularly useful for e.g. streaming compressed videos where a browser will send `Range: bytes=0-` then abort when satisfied, which is already hopeless for us to handle. At least when users scrub through video/audio, we can take a `Range: bytes=XXXXXX-` header and start the remote range request for the resource closer to the start of the byte range. This happens in increments of 1MB
- Add CRC32 checksum function for `CompressedResource` to make direct copying of compressed resources to new archives easier

~~Still getting `ERROR failed streaming asset error="resource: error 500: flate: corrupt input before offset 5"`, need to fix that before this becomes a proper PR because there's a bug somewhere~~


chocolatkey:
@copilot Can you find the bug that results in `ERROR failed streaming asset error="resource: error 500: flate: corrupt input before offset 5"` in the new changes in this PR?

chocolatkey:
All that needs to be done is using a readium repo for `gzran` (which will become `zran`)

PR #286, opened 2026-01-29
fix: allow Readium CLI to fetch files with spaces in name by resolving url for archivefetcher rather than straight .String()ing it

tefkah:
This PR aims to fix this issue in the Readium CLI: https://github.com/readium/cli/issues/94

I was not able to find a way to fix this purely on the CLI side, but my Go skills are very much lacking.

In short, this finds the `Path()`, which decodes possibly URLEncoded paths, rather than `String()` of the provided link to extract from the archive.

Otherwise, making requests to Readium cli for files with spaces in their titles fails, see the Readium CLI issue above.

I wasn't really able to write any tests that demonstrate this, as I'm not familiar enough with Go or the codebase! Sorry!

I did create a version of the test epub used in `archive` which has `s 04.xhtml` instead of `s04.xhtml`, in case that's of any use.
Also see the modified version of Moby Dick linked in the above issue for testing maybe!

this is [the archive test file, but renamed to .zip to allow uploading here](https://github.com/user-attachments/files/24937858/epub-with-space.zip)

see also this moby dick epub with spaces in the filenames. this is an epub contained in a zip 
https://github.com/user-attachments/files/24819589/Moby.Dick.Or.The.Whale.-.Herman.Melville.epub.zip

Disclosure: I mostly identified and solved this bug with the help of an LLM, specifically Claude Sonnet 4.5. The amount of code produced is so minimal
I feel comfortable sharing it unmodified, but I am totally unaware of any Go best practices etc to determine whether this way of fixing the bug makes sense or not!


pkg/fetcher/fetcher_archive.go
```diff
@@ -44,7 +44,12 @@ func (f *ArchiveFetcher) Links(ctx context.Context) (manifest.LinkList, error) {
 
 // Get implements Fetcher
 func (f *ArchiveFetcher) Get(ctx context.Context, link manifest.Link) Resource {
-	entry, err := f.archive.Entry(link.Href.String())
+	// use decoded path for archive entry lookup to support files with spaces and special characters
+	entryPath := link.Href.String()
```
chocolatkey: Please change to:
```go
var entryPath string
if hrefURL := link.Href.Resolve(nil, nil); hrefURL != nil {
    entryPath = hrefURL.Path()
} else {
    entryPath = link.Href.String()
}
```

chocolatkey (changes requested):
@tefkah sorry for the delay, please make the adjustment. Did you confirm that this fix addressed the problem for you? It's no 100% clear from the comment

tefkah:
oh sorry, yes, this does seem to fix it for me! ill apply that change asap

pkg/fetcher/fetcher_archive.go
tefkah: @chocolatkey sorry for the delay!

I want to reaffirm that ive tested running the readium-cli develop branch with this patch to work to resolve files with a space in them, while the  latest readium-cli without this patch does not

tefkah: and that ive pushed up the changes!

chocolatkey (approved):
Going to add some stuff after it's merged

PR #287, opened 2026-01-29
Use functions from `require`

### LINKED ISSUE #247: Use `require` instead of `assert` in tests
This is a low-priority issue. A lot of the tests use the [`assert` ](https://github.com/stretchr/testify?tab=readme-ov-file#assert-package) functions to check if tests are passing. For many of these, we should be using [`require`](https://github.com/stretchr/testify?tab=readme-ov-file#require-package) instead to stop execution of the test if, for example, there is an error during parsing, as there's no point in continuing with the test if an initial dependency fails

naglis:
Change assertions to utilize functions from `require` to stop tests immediately where subsequent assertions would make no sense.

Resolves #247


chocolatkey (approved):
@naglis thanks for doing this!

Issue #281, opened 2025-12-09
Prettier marshalling of a11y data


chocolatkey:
In the output, we can get for example:
```json
"accessibility": {
  "conformsTo": ["https://www.w3.org/TR/epub-a11y-11#wcag-2.2-aa"],
  "accessModeSufficient": [["textual"]],
}
```
These properties could be better marshalled and result in the following output:
```json
"accessibility": {
  "conformsTo": "https://www.w3.org/TR/epub-a11y-11#wcag-2.2-aa",
  "accessModeSufficient": ["textual"],
}
```

PR #290, opened 2026-02-03
Improve JSON marshaling of a11y metadata

naglis:
Addresses #281.


pkg/manifest/a11y.go
```diff
@@ -300,6 +309,17 @@ func A11yAccessModesFromStrings(strings []string) []A11yAccessMode {
 // A11yPrimaryAccessMode is a human primary sensory perceptual system or cognitive faculty through which a person may process or perceive information.
 type A11yPrimaryAccessMode string
 
+type A11yPrimaryAccessModeList [][]A11yPrimaryAccessMode
```
naglis, line 312: Not sure about the type name since it is a list of lists. Would welcome suggestions.

chocolatkey: @HadrienGardeur any suggestion for the name here?

HadrienGardeur: I would stick closely to the property name so something close from `accessMode` and `accessModeSufficient`, without too much additional fluff.

chocolatkey: @naglis could you make that adjustment? Let me know if you want an exact value

naglis: > Let me know if you want an exact value

I'd appreciate it 😅

